PR #18, opened 2021-03-20
Add create new tab feature to the modules generator application

Anonymxtrix:
Added a menu to the modules generator application and the create new tab feature.


MarcusTXK:
Okay looks good to me

MarcusTXK (commented):
Okay looks good to me

Issue #86, opened 2021-08-05
[RUNE]: Function arguments have 'any' type where they should not


houruomu:
Some of the functions in the Rune library do not have properly typed function arguments.

For example, the current ```color()``` function has signature ```color(rune: Rune, r: any, g: any, b: any): Rune```. However, it should instead be ```color(rune: Rune, r: number, g: number, b: number): Rune```.

This may cause confusion in the documentation.

See https://source-academy.github.io/modules/documentation/modules/runes.html#color for details.

houruomu:
Fixed in 35b987cc89cf84d0feefddbb739e914899597ce1

PR #91, opened 2021-08-17
[Bug Fix] stack_frac(), beside_frac(), overlay_frac() should allow frac in [0,1]

houruomu:
# Description
Previously ```stack_frac(), beside_frac(), overlay_frac()``` only took frac input in (0,1). They now correctly take frac input in [0,1].
*minor fix:* the input type of ```frac``` in these functions was ```any```, and is fixed to the correct type ```number``` now. See #86

# Known Issues 
Many of the input of other functions are still incorrectly indicated as ```any``` type, as described in #86 . I will fix this in later updates.


martin-henz (approved):
looks good to me

PR #92, opened 2021-08-18
[Rune] adding triangle

houruomu:
Bug: The "sail" rune was changed during the migration of rune library to modules. 
Fix: Added "triangle" rune, which is the same as the previous "sail" rune.


chownces (approved):
tested! LGTM thanks!

PR #96, opened 2021-08-21
[Improvement] improve the performance of hollusion

### LINKED ISSUE #95: [RUNE]: slowness when hollusion is run repeatedly
## Description
When `hollusion()` is run repeatedly, the browser starts to freeze and cpu consumption goes up.

## Example code
`hollusion(heart);`
Then click run several times
### Current behaviour
CPU consumption goes up and browser become slow
### Expected behaviour
Browser should not become slow

houruomu:
# Description
Students have reported that repeated run of `hollusion()` leads to memory leak/high memory consumption. This PR resolves the problem. Should close #95.

# Changes:
1. animation frames are buffered to alleviate CPU consumption
2. when code is re-run, previous animation is cancelled. This avoids `requestAnimationFrame()` building up in the background.


martin-henz (approved):
Looks fine!

PR #121, opened 2021-12-14
Add module for writing unit tests

podocarp:
# Description

Usually testing of programs in source academy is done by commenting and uncommenting code. This is a great big hassle. I have also noticed that students usually only run a single test at a time and comment/uncomment
every single line manually, which makes it even slower. Thus it would be great if we could introduce proper unit testing into Source.

This module adds a few function for writing tests and asserts, as well as a new tab to report on test statuses.
image

I have generally tried to keep to a style similar to `jest`. 

Future extensions:
- Compare runes, sound, etc.
- Style the tab a bit nicer

## Type of change

# How Has This Been Tested?

Added unit tests, as well as manual testing.

# Checklist:


modules.json
```diff
@@ -59,5 +59,10 @@
     "tabs": [
       "SoundMatrix"
     ]
+  },
+  "unittest": {
+    "tabs": [
+      "UnitTest"
+    ]
```
Anonymxtrix, line 66: Is there a reason we decided to name the module `unittest`? Don't think its an issue but I'm just thinking maybe just `test` is more straightforward and less confusing for our fellow freshmen.

src/bundles/unittest/asserts.ts
```diff
@@ -0,0 +1,128 @@
...
+        assert_contains(tail(xs), toContain);
+        return;
+      } catch (__) {
+        fail();
+      }
+    }
+  } else {
+    throw new Error(`First argument must be a list or a pair, got \`${xs}\`.`);
+  }
+}
```
Anonymxtrix, line 119: Try to avoid deep nesting as explained [here](https://nus-cs2103-ay2021s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-deep-nesting). This improves the readability of your program!

One method to do this is to use guard clauses instead of nesting if-else statements. 
Example:
```javascript
if (is_null(xs)) {
    fail();
}
// ...
```

src/tabs/UnitTest/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+type Props = {
+  result: any;
+};
+
+class UnitTests extends React.PureComponent<Props> {
+  /**
+   * Converts the results of a test suite run into a table format in its own div.
+   */
+  private static suiteResultToDiv(suiteResult: any) {
```
Anonymxtrix, line 17: Is it possible to restrict the type of `suiteResult` parameter here and try to move the type casting `... as SuiteResult` all the way to the point in which it is extracted from `this.props`? Reason is because we want to avoid dealing with `any` types as much as possible.

src/bundles/unittest/functions.ts
```diff
@@ -0,0 +1,82 @@
...
+  it: (msg: string, test: Test) => {
+    const name = `${msg}`;
+    let error = '';
+    context.suiteResults.total += 1;
+
+    try {
+      test();
+    } catch (err: any) {
+      error = handleErr(err);
+    }
```
Anonymxtrix, line 41: Is it possible to avoid mutation of the `error` variable here? Not really a huge problem in this function since scope of `error` is limited, but if function becomes bigger, it will start to become harder to keep track of all the mutations.

Anonymxtrix: There is also some mutations to other properties in this same `context` object which tbh I am uncomfortable with but I won't force u to change. But I would suggest minimising these mutations to reduce the number of bugs we may encounter in the future.

Anonymxtrix (commented):
Great job! I think that this test module that you have written will definitely be a hit among the source academy cadets! Too bad this wasn't yet implemented when I was taking 1101s haha.

modules.json
podocarp: I'll rename it into `testing`

src/bundles/unittest/asserts.ts
podocarp: Not sure how to simplify this

src/bundles/unittest/functions.ts
podocarp: `error` default to empty and is populated when an error is encoutered. I don't think this is a problem.

martin-henz:
This PR is still marked as "draft". @podocarp: Or is it ready for final review?

podocarp:
Have not written docs yet, but hope someone can look at the code first.

RichDom2185:
> I've looked through this PR, and as it stands it's basically done. There is an issue though: I don't see any way to have more than `describe` block per program.
> 
> Given that Source now supports multiple files and multiple calls to display results on tabs, if we don't mind removing `describe` I think I can get this to a working state.

I think the vast majority of use cases still involve single-file programs, so as long as it works for a single file, I think we can safely merge it first

RichDom2185 (approved):
LGTM based on @leeyi45's comment above, thanks!

PR #129, opened 2022-03-02
show functions should be identity

martin-henz:
now that the show function works as expected, it should return the given rune


martin-henz:
This is all done, as far as I can see. Closing this PR.

PR #141, opened 2022-05-17
Faster Rollup Rebuild Times by Selectively Building

Cloud7050:
This is the second pull request from the CSG module's fork. It only includes changes to the rollup build system (as opposed to Source Academy module changes), and aims to speed up the development/testing process of modules by greatly cutting down on time spent building after making incremental changes.

This PR brings about a single rollup config with some supporting files, while avoiding code duplication. It includes changes to related files such as `package.json` scripts and `.gitignore`, as well as to logging. The system works by storing a timestamp when a build or rebuild completes successfully. Subsequent rebuilds (`yarn rebuild`) then check for when each module's bundle/tab files were last modified, only building the modules that received changes, instead of having to build everything again or use other methods to limit what gets built (eg modifying `modules.json` or passing CLI arguments). The database is a simple json file powered by `lowdb` that is not meant to be committed similar to the `build/` folder, and can easily be deleted to force a complete build (or just `yarn build:rollup`).

image

Currently, this PR's listed changes are grouped together with the first PR's changes while it is still pending a merge. For easier comparing of changes for this PR alone, please refer to [this reference PR](https://github.com/Cloud7050/modules/pull/36/files).


leeyi45:
(Great if this functionality has already been implemented) Would it be possible to produce a json containing version or build information for each bundle? One of Prof Martin's other feature requests includes asynchronous module loading, which would be helped greatly if the build versions of the modules are made available

Cloud7050:
> (Great if this functionality has already been implemented) Would it be possible to produce a json containing version or build information for each bundle? One of Prof Martin's other feature requests includes asynchronous module loading, which would be helped greatly if the build versions of the modules are made available

It should be possible to use lowdb to repeatedly merge a modified version of `modules.json` into the build folder, as each rollup bundle is being run. This could contain individual timestamps for when that rollup bundle (be it a module bundle or module tab) was last built, for SA to base its decisions on. Then, the final plugin that updates the stored timestamp could further modify this json file instead of using its own like it does now (that would be the ending timestamp for when everything was more or less done building / being skipped).

Cloud7050:
image

PR #153, opened 2022-09-06
Improvements to the Module System 2

leeyi45:
This PR is a follow up to #148. motivated by the idea to streamline the existing module system.

Refer to [this](https://github.com/source-academy/frontend/pull/2208) and [this](https://github.com/source-academy/js-slang/pull/1289) for changes required to the frontend and js-slang.

# Primary Change
Currently, modules are written with the following format:
```.ts
// src/bundles/module/index.ts

import {
  // a bunch of things
} from './functions';

export default function module(moduleParams, moduleContexts) {
  // work with the module context
  
  return {
    // the module's exports
  }
}
```

## New Export Syntax
Under this new PR, modules are written following the ESM syntax
```.ts
// src/bundles/module/index.ts

export {
  // things to export
} from './functions';
```

If desired, modules can use a default export to export an object containing all its exports:
```.ts
// src/bundles/module/index.ts

export default new Proxy({
  // things to export
}, {
  get(target, name) {
    if (target[name] === undefined) {
      throw new Error(`unknown import: ${name}`);
    }
    return target[name];
  }
});

// OR more simply
import {
  // things to export
} from './functions';

export default {
  // things to export
};
```
Mixing default and named exports will cause the default export to become hidden, so this is not recommended.

## Accessing `context`
The `context` object is now available as an import:
```.ts
// src/bundles/curve/functions.ts
import { context } from 'js-slang/moduleHelpers';

const curvesDrawn = [];
context.moduleContexts.curve.state = {
  curvesDrawn,
};

// other module functions
```

`js-slang` doesn't actually export the context object, but a type declaration has been added for Rollup to convert to an external parameter when the bundles get transpiled to iife.

# Secondary Changes
## `moduleParams`
The `moduleParams` object has been removed since it was only being used by the `game` module, and the functionality it requires can be achieved using module contexts.

## Module Context Guarantees
In a source program, it is perfectly valid to have multiple imports on different lines to the same module so long as they are at the top level:
```.ts
import { draw_connected } from 'curve';

// some user code

import { unit_circle } from 'curve';
```

`js-slang` guarantees that no matter how many import calls to a module is made, each module file is only evaluated once per code evaluation, meaning that module states are only created once per evaluation.

This means that checking if the module context and state have been initialized is unnecessary:
```.ts
// Old code in curves/index.ts

// Update the module's global context
let moduleContext = moduleContexts.get('curve');

if (!moduleContext) {   
  moduleContext = {
    tabs: [],
    state: {
      drawnCurves,
    },
  };
  moduleContexts.set('curve', moduleContext);
} else if (!moduleContext.state) {
  moduleContext.state = {
    drawnCurves,
  };
} else {
  (moduleContext.state as CurveModuleState).drawnCurves = drawnCurves;
}
```
If your module code only initializes state information within module code, you can simply write:
```.ts
// curve/functions.ts
import { context } from 'js-slang/moduleHelpers';
const drawnCurves = [];

// The curve object on moduleContexts has already been created by js-slang
context.moduleContexts.curve.state = {
  drawnCurves,
}
```
For modules that initialize their state object outside of module code, the state object can be used in the following manner:
```.ts
// src/game/functions.ts
import { context } from 'js-slang/moduleHelpers';

const {
    scene,
    preloadImageMap,
    preloadSoundMap,
    preloadSpritesheetMap,
    remotePath,
    screenSize,
    createAward,
  } = context.moduleContexts.game.state || {};
```

Similarly, tabs can also follow the same guidelines:
```.tsx
// src/tabs/Curve/index.tsx
export default {
  toSpawn(context: DebuggerContext) {
   // If a module's bundle file is written properly, its module context should be properly initialized here
    const drawnCurves = context.context.moduleContexts.curve.state.drawnCurves;
    return drawnCurves.length > 0;
  },
  body(context: DebuggerContext) {
    // This is only called if toSpawn returns true, so can assume that this destructuring won't throw
    const { context: { moduleContexts: { curve: { state: { drawnCurves } } } } } = context;

    return <MultiItemDisplay elements={canvases} />;
  },
  label: 'Curves Tab',
  iconName: 'media', // See https://blueprintjs.com/docs/#icons for more options
};
```

## Build Script Changes
1. If you get the error that import assertions aren't supported, you may have to update your node version to the latest LTS
2. Because of the increasing size of bundles and their files, node may actually run out of memory space when running build tasks. Either run fewer tasks at once, or modify `package.json` with the `max-old-space-size` option as necessary: 
```.json
"scripts": {
    "scripts": "cross-env TS_NODE_PROJECT=\"scripts/tsconfig.json\" NODE_OPTIONS=\"--max-old-space-size=8192\" ts-node scripts/index.ts",
}
```


Cloud7050:
> Mixing default and named exports will cause the default export to become hidden, so this is not recommended.

Could you elaborate on this? Is this a result of the rollup process?

leeyi45:
Let's say you write a default export:
```.ts
export default {
  func1,
  func2,
}
```
or use named exports:
```ts
export {
  func1,
  func2,
} from './functions';
```
After Rollup when `js-slang` evaluates the bundle file, the module effectively becomes:
```ts
const __MODULE_0__ = { func1, func 2 };
```
If you mix default and named exports:
```ts
export default func1;
export {
  func2,
} from './functions';
```
The module gets transpiled as:
```ts
const __MODULE_0__ = { 
  func2,
  "default": func1
}
```
Right now `js-slang` doesn't support default imports, so you would have to do:
```ts
import { default } from 'module';
```
which wouldn't be valid source.
I think in the future we can start to allow namespace or default imports, but that functionality isn't available right now

martin-henz:
Does this PR work well with the currently deployed frontend in SourceAcademy@NUS? The reason why I'm asking is that we don't want to mess with the next modules being used in CS1101S. @leeyi45 @Cloud7050

Cloud7050:
> Does this PR work well with the currently deployed frontend in SourceAcademy@NUS? The reason why I'm asking is that we don't want to mess with the next modules being used in CS1101S. @leeyi45 @Cloud7050

It looks like the format in which built modules are provided will change since #138 (regarding IIFEs). This PR also looks like it involves changes from moduleParams/moduleContexts → context.moduleContexts. If NUS's frontend and js-slang versions are kept on an older deployed version for the purpose of stability, then the new format returned by the modules API may impact it.

Cloud7050:
Great that BlueprintJS has been updated here to match the frontend.

Going through the build process has helped me understand the new format of the built module bundles/tabs. However I seem to be unable to build the documentation.

> Docs Error: Failed to initialize Typedoc project

leeyi45:
> Does this PR work well with the currently deployed frontend in SourceAcademy@NUS? The reason why I'm asking is that we don't want to mess with the next modules being used in CS1101S. @leeyi45 @Cloud7050

The current master branch for the frontend is not compatible with the updated versions of modules and `js-slang` @martin-henz 

> Regarding building modules, I am quite surprised by the significant increase in build time compared to before, and am curious to know more about what has changed. It may be concerning, as developing modules can involve constant rebuilds.

@Cloud7050 I think it might have to do with the fact that `ts-node` needs to load all the files for the build scripts. Maybe it would be faster if we compiled the scripts. Not sure what the error regarding typedoc is, it works on both my Windows machine and M1 Mac. Ideally this is mitigated by the fact that running build will only build modified bundles and tabs.

Cloud7050:
(I edited my previous comment while you happened to be composing a reply. Please also refer to the added text.)

> ```
>   body(context: DebuggerContext) {
>     // This is only called if toSpawn returns true, so can assume that this destructuring won't throw
>     const { context: { moduleContexts: { curve: { state: { drawnCurves } } } } } = context;
> 
>     return <MultiItemDisplay elements={canvases} />;
>   },
> ```

I don't know if something has since changed, however it is worth pointing out bug source-academy/frontend#2136 where `body()` is called before `toSpawn()` is checked. If it has been fixed, then the issue can thankfully be closed.

I discovered this while working with the older version of module contexts, when developing CSG. I have a `Core` type that helped work with module states, and had to initialise it once on the bundle end and again on the tab end. I had to put the initialisation in `body()`, after which the initialised content was available to use in `toSpawn()`. If the bug has been fixed, I'll probably need to change that.

Cloud7050:
> This error arises if you're not using the right version of `js-slang`

I see, it it fitting for this PR to bump its js-slang version now that the associated Source PR has been merged?

> for the new module format to work, I needed to write an output plugin for Rollup that parses the bundles and tabs into an AST and make the edits there.

It's great that the string matching is no longer hardcoded on the receiving end. When I resume work on CSG, I will monitor the build times. As much as I prefer elegant solutions like dynamically removing the function call, the speed increase may be worth taking the "dumb" approach (I don't think the matching would need frequent/complex updating). Unless there are now other edits the AST method is making?

I am not sure how to proceed with the documentation issue. The error message doesn't reveal a lot. I've reinstalled packages, and my software has all been recently updated. For reference, I'm on Windows. Side note, is one of the "file modified" logs misfiring?

> ...
> • Csg: Build successful (1549.69 KB)
> 
> Docs Error: Failed to initialize Typedoc project
> Done in 247.56s.

> ... >yarn run build:docs
> yarn run v1.22.19
> $ yarn scripts build docs
> $ cross-env TS_NODE_PROJECT="scripts/tsconfig.json" NODE_OPTIONS="--max-old-space-size=8192" ts-node scripts/index.ts build docs
> (node:32880) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time
> (Use `node --trace-warnings ...` to show where the warning was created)
> • File modified: src\bundles\repeat\functions.ts
> Building HTML documentation
> 
> Building documentation for the following bundles:
> • repeat
> ...
> • csg
> Docs Error: Failed to initialize Typedoc project
> Done in 13.50s.

About ESLint, I'll probably take a look at the configs in detail down the line. If not now then I'll likely make some tweaks when I resume work on CSG.

leeyi45:
> > This error arises if you're not using the right version of `js-slang`
> 
> I see, it it fitting for this PR to bump its js-slang version now that the associated Source PR has been merged?
> 
> > for the new module format to work, I needed to write an output plugin for Rollup that parses the bundles and tabs into an AST and make the edits there.
> 
> It's great that the string matching is no longer hardcoded on the receiving end. When I resume work on CSG, I will monitor the build times. As much as I prefer elegant solutions like dynamically removing the function call, the speed increase may be worth taking the "dumb" approach (I don't think the matching would need frequent/complex updating). Unless there are now other edits the AST method is making?
> 
> I am not sure how to proceed with the documentation issue. The error message doesn't reveal a lot. I've reinstalled packages, and my software has all been recently updated. For reference, I'm on Windows. Side note, is one of the "file modified" logs misfiring?
> 
> > ...
> > • Csg: Build successful (1549.69 KB)
> > Docs Error: Failed to initialize Typedoc project
> > Done in 247.56s.
> 
> > ... >yarn run build:docs
> > yarn run v1.22.19
> > $ yarn scripts build docs
> > $ cross-env TS_NODE_PROJECT="scripts/tsconfig.json" NODE_OPTIONS="--max-old-space-size=8192" ts-node scripts/index.ts build docs
> > (node:32880) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time
> > (Use `node --trace-warnings ...` to show where the warning was created)
> > • File modified: src\bundles\repeat\functions.ts
> > Building HTML documentation
> > Building documentation for the following bundles:
> > • repeat
> > ...
> > • csg
> > Docs Error: Failed to initialize Typedoc project
> > Done in 13.50s.
> 
> About ESLint, I'll probably take a look at the configs in detail down the line. If not now then I'll likely make some tweaks when I resume work on CSG.
> 
> > It does not seem that Rollup is the bottleneck.
> 
> So from the earlier edit, do you think it's just a result of the extra work (AST) that rollup as a whole is now doing during the build process?

It wasn't the AST, actually that part is almost inconsequential in my testing. I'll mark this PR as a draft while I go do some benchmarking tests.

I usually build this stuff using my M1 MBP which takes me 2 mins tops with documentation, and that felt reasonable to me, so I never thought to try on another machine. Lesson learned, will definitely keep this in mind in the future.

leeyi45:
Building all bundles and tabs at the same time yielded the following timings on my Windows machine:
image

I used `rollup-timer` to time plugins, but the package hasn't been updated in a while and returns some crazy results
image

martin-henz:
Let's aim for merging and deploying for September 29. There is a week in CS1101S when modules are not used.
Till then, let's put this in "draft" mode.

Cloud7050:
I ran a few no-documentation builds on the new version. This was the fastest one:

image

The new `throw` from `buildDocsAndJsons()` means the full build script (with documentation) exits early and doesn't reach timings logging for me.

For comparison, I ran one full build on the previous rollup build system (around current master 46c7513). It took just `89.21s` *with* documentation, compared to this system's best time of `232.93s` *without* documentation, which I find is a highly significant difference. It was run on an overclocked i9-9900K, 32GB RAM, SSD. Maybe the difference in Mac vs Windows timings can be chalked up to Apple silicon, but when comparing the previous vs new systems on the same platform, I feel the difference warrants further investigation.

---

I noticed that edits to the `common/` tab Components alone aren't picked up as changes, probably because they aren't standalone tabs used in `modules.json`. This may require explicit `-m` when developing and testing them. Thought I'd bring it up, don't know if there's a more convenient way to have the script react to this.

---

As it looks like a window for merging is coming soon, I'll compile some of the earlier points:

- Increased timings
- Unknown documentation error
- Bump Source version?
- `body()` or `toSpawn()` called first as of now? May impact CSG's tab should it see use in Uppsala
- Common Component edits
- See also comment in https://github.com/source-academy/modules/pull/148#issuecomment-1250093707

PR #173, opened 2023-02-08
New Plotly Module

srj31:
# Description

Add a new module called `plotly`. It provides an interface to use the Plotly.js package to display graphs. 

Note: The plot will be used as a base for experiential learning for math concepts along with programming. Possible use cases to attempt: Matrix transformations, Taylor series, Three body simulations, etc.

The modal button isn't styled yet, will be worked on later.

## Type of change

# How Has This Been Tested?

It currently only exposes the `new_plot` function which requires a list of pair of fields and value for the data object that the plotly.js API uses to construct the plot.

Given the correct pairing of fields and value of the objects the plot can be seen on the Plotly tab.

# Checklist:


leeyi45:
Could you provide some more documentation as to how `new_plot()` is to be used? Right now the example you've given is a bit mysterious....

martin-henz:
sourceacademy.org has a "Full JavaScript" option. For that, it would be nice to have a JSON option for the input format, instead of insisting on the list-and-array format of Source §x (no objects).

It would be super-nice, if the user could choose among the two data formats. (We could even allow the user to freely mix the formats, if that doesn't complicate things too much.)

srj31:
- Added a more information about the type of data new_plot supports
- Added new_plot_json which would use the entire plotly library using json objects in full Javascript mode.

martin-henz (approved):
Looks good. Let's play with this and see how we can improve the teaching material. Replacement curves (or even runes) seems possible.

PR #177, opened 2023-02-09
Programmable Repl module source codes with config

larrywang0701:
Important notes about the required node modules:

  In package.json / dependencies:

    The two node modules "ace-builds" and "react-ace" are necessary and (for now) only used by the Programmable Repl module.
    In react-ace version 10.1.0, ace-builds must be version 1.4.14, NOT the latest version. The latest version of ace-builds will cause compile error in react-ace due to compatibility issues.
    Also, just make the react-ace version fixed to 10.1.0 to prevent more compatibility issues between ace-builds and react-ace.

    So please DON'T add the "^" symbol before the version string for "ace-builds" and "react-ace" in package.json / dependencies.
    And also DON'T edit the part in yarn.lock file related to the above two node modules.
    It really cost me some time to fix the compatibility issue up with packages.json and yarn.lock .
    If you modify the related part in these two files, your code may can't pass typechecking when building.
    Thanks!

    --Wang Zihan

### DIFF
--- a/.github/workflows/pages-deploy.yml
+++ b/.github/workflows/pages-deploy.yml
@@ -23,6 +23,7 @@ jobs:
           
       - name: Build Modules 🔧
         run: |
+          mkdir build
           yarn run build
 
       - name: Deploy 🚀


leeyi45:
Hi, please provide some documentation as to how we can use your module

leeyi45:
Just wondering if we want to keep this module's name as `programmable_repl`. It seems a bit long.

src/bundles/programmable_repl/functions.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+
+/**
+ * Redirects the display message into Programmable REPL Tab and returns the string parameter
+ * @param {content} the content you want to display
+ *
+ * @category Main
+ */
+export function module_display(content : any) : any {
```
leeyi45, line 37: This may be better named as `display_in_repl` or something to that effect. Our convention is to name methods with verbs, and `module_display` doesn't quite convey what this method is used for

src/tabs/ProgrammableREPL/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+import 'ace-builds/src-noconflict/mode-javascript';
+import 'ace-builds/src-noconflict/theme-twilight';
+import 'ace-builds/src-noconflict/ext-language_tools';
+
+
+type GUI_Data = {
+  programmableReplInstance: ProgrammableRepl;
+};
+
+class ProgrammableReplGUI extends React.Component<GUI_Data> {
```
leeyi45, line 22: The convention is to name it `***Props`, but it's alright, just need to change the casing.

src/tabs/ProgrammableREPL/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+            width: '100%',
+            height: '375px',
+            ...(this.replInstance.customizedEditorProps.backgroundImageUrl != 'no-background-image' && {
+              backgroundImage: `url(${this.replInstance.customizedEditorProps.backgroundImageUrl})`,
+              backgroundColor: `rgba(20, 20, 20, ${this.replInstance.customizedEditorProps.backgroundColorAlpha})`,
+              backgroundSize: '100%',
+              backgroundRepeat: 'no-repeat',
+            }),
+          } }
+          mode="javascript" theme="twilight"
```
leeyi45, line 64: Improvements in the future to consider? Maybe let users set the theme value for themselves? That would be quite cool imo.

src/bundles/programmable_repl/programmable_repl.ts
```diff
@@ -0,0 +1,202 @@
...
+  	this.outputStrings = [];
+    let retVal : any;
+    try {
+      if (this.evalFunction === default_js_slang) {
+        retVal = this.evalFunction(this.userCodeInEditor, document.body);
+      } else {
+        retVal = this.evalFunction(this.userCodeInEditor);
+      }
+    } catch (exception : any) {
+    	this.OutputString(`Line ${exception.location.start.line.toString()}: ${exception.error.message}`, 'red');
```
leeyi45, line 68: As we've discussed, `error` may not always have a `location` property

src/bundles/programmable_repl/programmable_repl.ts
```diff
@@ -0,0 +1,202 @@
...
+  	}
+  	let tmp = {
+  		content: str,
+  		color: textColor,
+  	};
+  	this.outputStrings.push(tmp);
+  }
+
+  // Returns the forbidden word present in the string "str" if it contains at least one unsafe word. Returns "safe" if the string is considered to be safe to output directly into innerHTML.
+  private OutputStringSafeCheck(str) {
```
leeyi45, line 102: We've discussed how this is XSS vulnerable, just putting this here for documentation.

src/bundles/programmable_repl/programmable_repl.ts
```diff
@@ -0,0 +1,202 @@
...
+    const prelude = 'const display=(x)=>module_display(x);';
+    (context as any).sourceRunner(prelude, jsslangContext, true, options)
+      .then((preludeEvalResult) => {
+        developmentLog(preludeEvalResult);
+        if (preludeEvalResult.status === 'error') {
+          this.OutputString('[Warning] It seems that you havn\'t import the function "module_display" correctly when calling "invoke_repl". The runner will use the default js-slang builtin display function.', 'yellow');
+        }
+      });
+    let result : Promise<any> = (context as any).sourceRunner(code, jsslangContext, true, options);
+    developmentLog(result);
```
leeyi45, line 147: Remember to remove these when you're done!

src/bundles/programmable_repl/programmable_repl.ts
```diff
@@ -0,0 +1,202 @@
...
+    result.then((evalResult) => {
+      if (evalResult.status != 'error') {
+        this.OutputString('js-slang program finished with value:', 'cyan');
+        this.OutputString(evalResult.value === undefined ? 'undefined' : evalResult.value.toString(), 'cyan');
+      } else {
+        const errors = jsslangContext.errors;
+        const errorCount = errors.length;
+        for (let i = 0; i < errorCount; i++) {
+          const error = errors[i];
+          this.OutputString(`Line ${error.location.start.line}: ${error.type} Error: ${error.explain()}  (${error.elaborate()})`, 'red');
```
leeyi45, line 157: Here also has the error location nullability problem

leeyi45 (commented):
Great work, just some changes needed

martin-henz:
How about just calling the module "repl"?

PR #187, opened 2023-02-21
Fix critical security issue, adjust function names

larrywang0701:
Fix the critical security issue that the safe-check function is only applied on the first character of rich text string (not the whole string).
Use shorter student-side function names as Martin asked, and merged "rich_display" into "module_display"


larrywang0701:
Github check my code as unsuccessful with given
```
Verify all tests pass and build successThis is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on April 1st, 2023. For more details, see https://github.com/actions/runner-images/issues/6002Show less
 ```
Seems like a Github internal error that is not related to my own code.

martin-henz (approved):
Nice! Looking forward to using the repl module in CS1101S and CS4215.

PR #193, opened 2023-04-02
Plotly curve sound enhancements

srj31:
# Description

Provides visualization enhancements to the curves module, allowing the use of interactive plotly graphs for drawing curves in source academy as well as visualizing sounds to make the sound modules more accessible to people and provide a different insight and feedback for sound module assignments.

## Type of change

# How Has This Been Tested?

The visualizations were tested with the existing assignments and the features worked as expected without requiring any changes to the initial working logic. 

# Checklist:


src/bundles/plotly/sound_functions.ts
```diff
@@ -0,0 +1,99 @@
...
+    let temp: number;
+    let prev_value = 0;
+
+    const wave = get_wave(sound);
+    for (let i = 0; i < len; i += 1) {
+      temp = wave(i / FS);
+      time_stamps[i] = i / FS;
+      // clip amplitude
+      // channel[i] = temp > 1 ? 1 : temp < -1 ? -1 : temp;
+      if (temp > 1) {
```
martin-henz, line 41: I think amplitude should not be clipped for draw_sound. That way if clipping happens, one can use this tool to see the original wave function.

src/bundles/plotly/sound_functions.ts
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Visualizes the sound on a 2d line graph
+ * @param sound the sound which is to be visualized on plotly
+ */
+export function draw_sound_2d(sound: Sound) {
+  if (!is_sound(sound)) {
+    throw new Error(`play is expecting sound, but encountered ${sound}`);
+    // If a sound is already playing, terminate execution.
+  } else if (get_duration(sound) < 0) {
+    throw new Error('play: duration of sound is negative');
```
martin-henz, line 23: should be "draw_sound_2d"?

src/bundles/plotly/sound_functions.ts
```diff
@@ -0,0 +1,99 @@
...
+  drawnPlots,
+};
+
+/**
+ * Visualizes the sound on a 2d line graph
+ * @param sound the sound which is to be visualized on plotly
+ */
+export function draw_sound_2d(sound: Sound) {
+  if (!is_sound(sound)) {
+    throw new Error(`play is expecting sound, but encountered ${sound}`);
```
martin-henz, line 20: should be "draw_sound_2d is..."?

src/bundles/plotly/sound_functions.ts
```diff
@@ -0,0 +1,99 @@
...
+      // channel[i] = temp > 1 ? 1 : temp < -1 ? -1 : temp;
+      if (temp > 1) {
+        channel[i] = 1;
+      } else if (temp < -1) {
+        channel[i] = -1;
+      } else {
+        channel[i] = temp;
+      }
+
+      // smoothen out sudden cut-outs
```
martin-henz, line 49: do we need this?

martin-henz (changes requested):

src/bundles/plotly/sound_functions.ts
srj31: Clipping was done in this case as well, to prevent the graph from getting too zoomed out due to an unnecessarily large output for one of the timestamps since plotly will initially want to show all the data points. That being said your point is correct, since it will be a good way of seeing why was the audio not the way the user expected it to be.

martin-henz (approved):

Issue #196, opened 2023-04-03
[Feature Request]: Modularise remote execution


RichDom2185:
# Feature

Related to https://github.com/source-academy/frontend/issues/2412.

> As we support more and more types of devices (e.g. SPIKE Prime, Arduino), it would be good if the logic for the different modes of execution for the different devices be separated away from the frontend.

Let us migrate the logic for remote execution to the modules repository.

## Checklist

- [x] I checked the requested feature does not already exist
- [x] I checked to make sure that this issue has not already been filed

PR #197, opened 2023-04-03
Add remote execution module

RichDom2185:
Creates a new module, `remote_execution`, and populates it with the EV3's internal functions as an initial version.

Content was copied over from https://github.com/source-academy/frontend/blob/8ffd98040d28c0090c12b02fb36b9c7eb02e2f51/src/features/remoteExecution/RemoteExecutionTypes.ts#L52.

Related to #196.

## Type of change


martin-henz (approved):
I agree that we should put as much of the remote execution functionality into modules as we can.

PR #199, opened 2023-04-05
Unity Academy Update

larrywang0701:
Here is an update for my Unity Academy module
It may not work now, because I haven't deploy the new version of Unity Academy frontend app that corresponding to this update to the backend
I'll deploy the new Unity Academy frontend to the backend only after this PR is merged, because if I deploy the new version of Unity Academy frontend now, the current Unity Academy module (old version) will no longer work.

### DIFF
--- a/src/bundles/unity_academy/UnityAcademy.tsx
+++ b/src/bundles/unity_academy/UnityAcademy.tsx
@@ -586,12 +586,12 @@ class UnityAcademyJsInteropContext {
     this.targetFrameRate = newTargetFrameRate;
   }
 
-  setCustomPropertyInternal(gameObjectIdentifier : GameObjectIdentifier, propName : string, value : any) : void{
+  setCustomPropertyInternal(gameObjectIdentifier : GameObjectIdentifier, propName : string, value : any) : void {
     const gameObject = this.getStudentGameObject(gameObjectIdentifier);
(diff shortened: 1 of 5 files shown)


src/bundles/unity_academy/UnityAcademy.tsx
```diff
@@ -542,13 +579,23 @@ class UnityAcademyJsInteropContext {
...
 
-  setTargetFrameRate(newTargetFrameRate : number) {
+  setTargetFrameRate(newTargetFrameRate : number) : void {
     newTargetFrameRate = Math.floor(newTargetFrameRate);
     if (newTargetFrameRate < 15) return;
     if (newTargetFrameRate > 120) return;
     this.targetFrameRate = newTargetFrameRate;
   }
 
+  setCustomPropertyInternal(gameObjectIdentifier : GameObjectIdentifier, propName : string, value : any) : void {
```
martin-henz, line 589: @leeyi45 raised concerns about the propName parameter: he was thinking there was a XSS vulnerability here. I think that's not the case, because the entered string is going to remain uninterpreted as a string at runtime. It cannot escape being a string: The string is not going to be subjected to any language processing (e.g. compilation).

martin-henz (approved):

PR #202, opened 2023-04-10
Arcade2D Module

tituschewxj:
# Description

Adds a new module called Arcade2D that supports animation of 2D drawings, sprite images, keyboard and mouse input, and audio output. The intended users of this module are CS1101S students who have completed up to Source 3. Arcade2D is a wrapper for the Phaser 3, and aims to simplify and streamline the implementation and allow students to make games using code similar to that they have already learnt.

## Type of change

# How Has This Been Tested?

We tested this by creating a demo game, and other examples and the features work as expected.
Some examples are documented in the index file.

# Checklist:


scripts/src/templates/module.ts
larrywang0701, line 1: It seems that your fork or local development enviroment is not synced properly with the main repo. You are trying to commit to these files related to the overall module building system, and I don't think an individual module should modify these files without a very well-described and detailed reason. Try not commit to these files if you not intend to do so.

src/bundles/arcade_2d/index.ts
larrywang0701, line 1: Personally I think better with more examples? Since you know your module and every functions inside very well, but for new users this may not. A concept that may looks very simple and basic to you, but may be very unfamiliar to people who just learn game programming, or using your module. As the author of Unity Academy module, I can understand what's "main loop" and "game objects" are, but for new programmers, they may even don't know about the fact that "video games is a big loop". Same thing happened to my Unity Academy module before when my friends are trying to understand my documentation. So maybe try add more examples and documentations to help others better understand your module?

larrywang0701 (changes requested):
Please try not touch the module build system files, or please provide a very detailed reason on why you need to modify the overall module build system files. If you can't 100% know what you are doing, this may cause errors and side effects in other modules.

larrywang0701 (commented):
Please confirm on that is it necessarily to change the overall shared module build system files, and please self-check on the security and performance stuffs (for a game engine I think it would lead to a waste of resources (CPU, GPU, battery, etc) on user's devices if it just simply always run in background until user totally close the browser page. In Unity Academy users can choose to manually terminate the engine, and Unity Academy has auto sleep design to reduce resource cost. But since pixnfilx don't have such design, I'm not really sure about the situation in your module here.

scripts/src/templates/utilities.ts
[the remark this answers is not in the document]
tituschewxj: Because when I tried to name it without the edit I get the warning below. I'm not sure how you were able to name it with it being valid. I don't think this will affect other modules, since the function `isSnakeCase` is called only when a module is created and this change only expands the possible set of bundle names. I have tested the yarn scripts on this module and doesn't seem to be any problem arising from this change in definition for snake case in the module building process.
Screenshot 2023-04-12 211144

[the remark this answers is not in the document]
larrywang0701: Try directly name it in the json file and path. You don't need to run `yarn create` again after you already created it. Actually I didn't use the `yarn create` command to create the module, and you can also see that the name `physics2d` is also a valid name without changing the file (refer to their PR). Personally I think this modification would unnecessary, if you think this is OK, I can approve it but remember please always be responsible to your code.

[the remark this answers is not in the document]
tituschewxj: Directly renaming the module is an option. However, I feel it is inconsistent if there are module names that violate the snake case naming convention. Suppose there are other modules that want to include numbers in their names in future (following our modules), they might find the same problem of being unable to do so unless they directly rename (which defeats the purpose of not modifying the snake case definition and gives extra work to the creators of that new module)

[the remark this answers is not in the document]
larrywang0701: OK, I think you do have a point, and just allowing numbers in module name seems has no effect on the current modules. I'll approve for you now, and after you self-check on the security stuffs I mentioned (especially on the part that students need to provide an arbitrary string or a function object as a parameter, like your `create_sprite` or `create_text` functions.) and confirmed that all things are ok, you can merge it by yourself. Please remember always be responsible to your code and the users when you do that. Thanks :)

(An injection could happen when attackers intentionally put some JS code as strings into your vulnerble functions' parametrs to let the VM interpret it and execute as JS code, and for most freshman in CS1101S I believe they won't check the code from Source Academy's share links before they press Run button)

Also, a small additional advice here, you may consider this in the future to better help users to use your module. Maybe you can consider adding a parameter type check using the javascript keywork typeof. For example, what will happen if students provide wrong parameter types like `create_circle("123")`? A small improvement could be done to help students use the module better :)

larrywang0701 (approved):
You can merge it by yourself if you finished self-checking and think all things are safe and well.

src/bundles/arcade_2d/functions.ts
```diff
@@ -0,0 +1,891 @@
...
+
+// =============================================================================
+// Global Variables
+// =============================================================================
+
+let WIDTH: number = DEFAULT_WIDTH;
+let HEIGHT: number = DEFAULT_HEIGHT;
+let SCALE: number = DEFAULT_SCALE;
+let FPS: number = DEFAULT_FPS;
+export let DEBUG: boolean = false;
```
RichDom2185, line 75: Minor nitpick: perhaps, instead of leaving a bunch of "naked" global variables, it would be better to encapsulate this information in some sort of settings object.

src/tabs/ArcadeTwod/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+};
+
+/**
+ * Component for UI buttons within tab e.g play/pause.
+ */
+class A2dUiButtons extends React.Component<UiProps, UiState> {
+  constructor(props) {
+    super(props);
+    this.state = {
+      paused: false,
```
RichDom2185, line 57: Boolean variables should ideally have names that show they are boolean for readability.

```suggestion
      isPaused: false,
```

RichDom2185 (changes requested):
Hi, great work on the module! Was looking through some of the PRs and leaving my comments as below, do take a look at them at let me know what you think!

Thanks for the hard work!

src/bundles/arcade_2d/types.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Represents the render properties of a GameObject.
+ * @property {Color} color - The color associated with the GameObject tint.
+ * @property {[number, number]} flip - The (x,y) flip state of the GameObject.
+ * @property {boolean} visible - The render-visibility of the GameObject.
+ */
+export type RenderProps = {
+  color: Color;
+  flip: [boolean, boolean];
+  visible: boolean;
+};
```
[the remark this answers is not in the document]
RichDom2185: Same thing for the `[boolean, boolean]`.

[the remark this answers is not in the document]
xenosf: Thank you for the suggestion. I've changed the code to use type aliases.

src/bundles/arcade_2d/functions.ts
tituschewxj: Ok, thanks for the suggestion. I have encapsulated the information in a config object.

src/bundles/arcade_2d/phaserScene.ts
```diff
@@ -0,0 +1,358 @@
...
+/**
+ * The Phaser scene that parses the GameObjects and update loop created by the user,
+ * into Phaser GameObjects, and Phaser updates.
+ */
+export class PhaserScene extends Phaser.Scene {
+  constructor() {
+    super('PhaserScene');
+  }
+  private sourceGameObjects;
+  private phaserGameObjects = [] as (PhaserGameObject)[];
```
RichDom2185, line 56: I'm not sure why the type cast here is necessary. Did you mean:

```suggestion
  private phaserGameObjects: Array<PhaserGameObject> = [];
```

src/bundles/arcade_2d/gameobject.ts
```diff
@@ -0,0 +1,387 @@
...
+    this.renderNotUpdated = true;
+  }
+  hasRenderUpdates(): boolean {
+    return this.renderNotUpdated;
+  }
+  updatedRender() {
+    this.renderNotUpdated = false;
+    this.shouldBringToTop = false;
+  }
+  getColor(): [number, number, number, number] {
```
RichDom2185, line 105: Perhaps this should be using the `ColorRGBA` type alias instead?

src/bundles/arcade_2d/audio.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Encapsulates the representation of AudioClips.
+ * AudioClips are unique - there are no AudioClips with the same URL.
+ */
+export class AudioClip {
+  private static audioClipCount: number = 0;
+  // Stores AudioClip index with the URL as a unique key.
+  private static audioClips: Map<string, number> = new Map<string, number>();
+  // storea all the created AudioClips
```
RichDom2185, line 13:
```suggestion
  // Stores all the created AudioClips
```

RichDom2185 (commented):
Thanks for working on the changes! There are some more minor comments as below. No rush to do them, please focus on your exams first.

A not-so-minor nitpick, while it's good that the boolean state variable `paused` has been changed to `isPaused`, there are still quite a few more booleans/functions that return booleans that are not following the usual naming convention of `isXXX`, `hasXXX`, `shouldXXX`, etc., and this affects readability.

For example, under `audio.ts`, possible changes may be (feel free to change accordingly if you feel that a better name captures the purpose of the variable more effectively):

* `loop` &rarr; `shouldLoopAudio`
* `getLoop()` &rarr; `isLooping()`
* `playClip` &rarr; ...
* `loopAudioClip()` &rarr; `setShouldLoopAudio()`
* ...

Note the naming of the setters (generally starting with "set").

Thanks for the hard work!

martin-henz:
What's the status of this PR?

tituschewxj:
The status of this PR is that there are some minor changes to do from the reviewers' comments. We should be ready to merge once we have resolved the reviewers' comments and suggestions.

src/bundles/arcade_2d/gameobject.ts
tituschewxj: Thanks for all the help!

src/bundles/arcade_2d/audio.ts
```diff
@@ -35,43 +35,47 @@ export class AudioClip {
...
   }
   /**
    * Checks if the Audio Clip needs to update. Updates the flag if true.
    */
   public hasAudioClipUpdates() {
-    const temp = this.audioClipNotUpdated;
-    this.updatedAudioClip();
+    const temp = !this.isUpdated;
+    this.setAudioClipUpdated();
     return temp;
```
RichDom2185: Minor nit. Since this is not exactly a "temporary" variable (the value is returned, after all), perhaps it would make more sense to call this `prevValue`?

```suggestion
    const prevValue = !this.isUpdated;
    this.setAudioClipUpdated();
    return prevValue;
```

RichDom2185 (commented):
Just one minor nitpick. Great work guys! I'll leave it to the other reviewers for further comments.

larrywang0701 (approved):
If there's no more problems in this PR, then I think it can be merged. You can merge it by yourself if you think everything is ok and ready

larrywang0701 (approved):

RichDom2185 (approved):
Thanks, LGTM! Feel free to post some sample programs (via a shared link) to demonstrate these new additions :)

Thanks for the hard work!

PR #204, opened 2023-04-11
WebAssembly Module

Yongbeom-Kim:
# Description

Webassembly module for SA

## Type of change

# How Has This Been Tested?

Tested in separate repo, [here](https://github.com/Yongbeom-Kim/wabt-parser)

# Checklist:


src/bundles/wasm/index.ts
RichDom2185, line 1: Could you change the line-encoding of this file to CRLF instead of LF, in order to match the rest of the codebase?

It's a bit weird, but this repository uses CRLF encoding for files, even on Unix systems.

To change the line encoding, open the file on VSCode, the on the bottom right toolbar:

image

Click _"Select End of Line Sequence"_ and change it to CRLF, then commit the changes.

RichDom2185 (commented):
Hi! Just a minor comment, thanks for the hard work, looks good!

src/bundles/wasm/index.ts
Yongbeom-Kim: Thanks for the review! 

I've changed the relevant settings on VS Code and fixed all eslint issues, which should have fixed the problem? I'm not entirely sure if the problem is gone, but eslint isn't giving me any warnings so I take it that the issue is fixed. 

Please let me know if there's anything else (or if the issue persists)!

tituschewxj: I don't think the line-encoding is updated in the GitHub repository, it is only updated on your local files. You can check if there are warnings by refering to the GitHub Checks.
Screenshot 2023-04-14 141923
I think your `git config core.autocrlf` is set to true, so git will replace CRLF with the line encoding in the GitHub repository automatically, which is LF. Which is why it is not updated on the GitHub repository, even if you change it locally.

I faced a similar problem with the line-encodings, but I was able to fix it by referring to this: https://stackoverflow.com/questions/1967370/git-replacing-lf-with-crlf

leeyi45: At some point someone configured the repository to use CRLF instead of LF (I believe it was @Cloud7050) within the eslint configuration, I've since left that as is. If we want to move back to LF it would be some minor changes to the build system

Cloud7050: Using `git ls-files --eol`, this repo appears to sometimes use LF and sometimes use CRLF in the index (and some files are even mixed). For comparison, the main js-slang repo appears to _mainly_ use LF.

This repo's base eslint config is based on a config I've been carrying with me for a while. As I'm on Windows, that rule's setting is less of a concious choice for this repo. If I'm understanding this correctly, maybe it'll be better to:

- Rely on each user's git `core.autocrlf` config to handle line endings across platforms (true on Windows, input on Linux/macOS?)
- Turn off the `linebreak-style` rule in the base eslint config. Users are free to locally checkout LF or CRLF
- Push a commit that converts all existing files to be indexed with LF (could use something like [eol-converter-cli](https://www.npmjs.com/package/eol-converter-cli) on npm)

From there I think the repo may sometimes get CRLF introduced, but they would sometimes get corrected to LF down the line. Any thoughts?

Yongbeom-Kim: Thanks for the help with the line-encoding issue - I've fixed the issue with this PR, let me move this discussion to a fresh issue. #209

martin-henz: Is this PR ready to be merged?

RichDom2185:
I think we can merge this module in. @Yongbeom-Kim could you kindly resolve the conflicts? Thanks!

Yongbeom-Kim:
Sorry for the late follow-up on this, PR is ready to be merged!

RichDom2185 (approved):
LGTM, thanks!

PR #206, opened 2023-04-12
CSG Module 2.0

joeng03:
# Description

Objectives:

Statement:
Fix the current issue of incorrect display of colour after boolean operations (union, subtract, intersect) on primitive shapes. Currently, the colours of the initial Shapes are not preserved in the resulting Shape, instead a default colour (grey) is used. 

Approach: 
This can be achieved through patching the @jscad/modeling package. We have used patch-package to ensure that our patch is applied seamlessly every time a new installation (i.e. every run of yarn install)  is performed. By tracing the code and referring to Version 1 of JSCAD and CSG.js, we found out that the colors are not preserved due to the single polygons which make up a solid not storing the color. The boolean operations are carried out through a Binary Space Partitioning (BSP) tree data structure, in which the color of polygons are not preserved during when they are split. We modified several methods to ensure that the colors of individual Shapes are stored in the polygons after boolean operations. Since retesellation combines coplanar planes even when they have different colors, we prefer to remove the retesellation functionality for this stage of development.

Statement: 
Introduce hierarchy of Shapes by adding Groups. Groups can consist of Shapes or Groups. When transformations are applied to Groups, all objects in the Group undergo the transformation. 

Approach: 
A new class, Group was added to the utilities.ts file. Group takes in a List of Group|Shape, and contains an attribute, transforms, which stores the transformation matrix of the Group. 
Both the Shape and Group classes now implement the Entity interface, which contains the transformation methods.

Statement: 
Delay transformations on objects until the object is to be rendered to increase speed during modelling. Combine transformations when performing the final transformation, so that each individual Shape only undergoes one transformation operation. This feature has to integrate with Groups. 

Approach: 
Both the Shape and Group classes now implement the Entity interface, which contains the transformation methods. Due to considerations for efficiency, Groups actually shallow copies all its children under the hood during transformations, while creating an impression of immutability (new object created) on the client side. Deep copying is done lazily (delayed until rendering stage). During the rendering stage, the transformations are passed down the tree structure of the Group.

Statement: 
Enforce the immutable and side-effect-free functional programming experience of the CSG module. This includes enforcing the immutability of shapes, the creation of a new object for every operation, and the removal of statefulness during rendering.

Approach:
Every primitive is now exported as a function which takes in a hex string as color, and further mutation of colors of a Shape throughout its lifecycle are now disallowed. Furthermore, the store() function is removed, and its functionality is now achieved by passing either a Group or a Shape into the render() function in the final lines of the code.

Statement: 
Add a STL serializer to serialize CSG shapes into .stl files, for the purpose of 3D printing.

Approach:
Add another dependency, @jscad/stl-serializer, which exports a method, serialize, to convert a JSCAD shape into an .stl file.          

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


patches/@jscad+modeling+2.11.0.patch
RichDom2185, line 1: Did you intend to commit this file change? If not, could you remove it?

.idea/.gitignore
RichDom2185, line 1: I think the files in the `.idea` folder may have been mistakenly committed into your branch. Do you mind deleting them?

RichDom2185 (commented):
Hi, taking a quick glance at the PRs and noticed that you may have accidentally committed some files as below, could you check again?

Thanks for the hard work!

.idea/.gitignore
joeng03: just removed!

martin-henz:
What's the status of this PR? Merge?

joeng03:
It is still under development as we still have several bugs to resolve. Now that finals has ended, I will work on it with Chenbo.

RichDom2185:
> It is still under development as we still have several bugs to resolve. Now that finals has ended, I will work on it with Chenbo.

Ok, in that case, let us mark it as draft first. Feel free to mark the PR as ready when the bugs have been resolved!

joeng03:
Hi @RichDom2185, we have resolved the bugs and ready to merge!

RichDom2185:
> Hi @RichDom2185, we have resolved the bugs and are ready to merge!

Do you have some demo/example code to play around with this module?

src/bundles/csg/functions.ts
```diff
@@ -10,90 +10,198 @@
...
+ *
+ * @param {Shape} shape - The Shape to be coloured and returned
+ * @param {string} hex - The colour code to use
+ * @returns {Shape} The colorized shape
+ */
+function colorize(shape: Shape, hex: string) {
+  let color: Color = hexToColor(hex);
+  let coloredSolid: Solid = _colorize(color, shape.solid);
+  return new Shape(coloredSolid);
+}
```
RichDom2185, line 50: Just like the workflow warning says, this function does not seem to be used anywhere?

src/bundles/csg/functions.ts
```diff
@@ -315,324 +486,75 @@ export function intersect(a: Shape, b: Shape): Shape {
 }
 
 /**
- * Scales the shape in the x, y and z direction with the specified factor,
- * ranging from 0 to infinity.
- * For example scaling the shape by 1 in x, y and z direction results in
+ * Scales the shape in the x, y and z direction with the specified factor.
+ * Factors must be non-zero.
```
RichDom2185, line 490: Why must factors be non-zero? Strictly speaking, shouldn't it be technically possible to scale an entity by a factor of 0 (making it infinitesimally small)?

If it is to get around an error, I don't think this is the best way to do it.

RichDom2185 (commented):
Was there any reason functions like `scale_(x|y|z)`, `flip_(x|y|z)`, `transform_(x|y|z)`, `shape_center`, `area`, `volume`, etc. were removed from the list of exports?

src/bundles/csg/functions.ts
joeng03: The JSCAD code (node_modules\@jscad\modeling\src\operations\transforms\scale.js) currently does not even allow negative values and will throw an error if factors are <=0. We decided to add a <=0 check in function.ts to avoid revealing the JSCAD implementation

src/bundles/csg/functions.ts
joeng03: > Was there any reason functions like `scale_(x|y|z)`, `flip_(x|y|z)`, `transform_(x|y|z)`, `shape_center`, `area`, `volume`, etc. were removed from the list of exports?
It was a change requested by Prof Low to make the set of exported functions less redundant

> Just like the workflow warning says, this function does not seem to be used anywhere?
Thanks for reminding :) Going to remove it in the latest commit

joeng03:
> > Hi @RichDom2185, we have resolved the bugs and are ready to merge!
> 
> Do you have some demo/example code to play around with this module?

Its in samples.md, do let me know if you run into any issues with the sample code!

src/bundles/csg/functions.ts
RichDom2185: This sounds like we are artificially enforcing that factors are positive simply to get around this limitation/error of the JSCAD library. It's not about revealing implementation; what I meant was more of a question of "should we support scaling by non-positive factors?".

If we can define scaling by a negative number to be _"flip in the corresponding axes that are scaled by a negative factor, and then scale by the absolute value of the factors" then why can't we just implement it that way in the code? Artificially capping it to >= 0 just because it will through an error otherwise does not seem like the best thing to do. Is there any reason (perhaps pedagogical/etc.) that it was not done this way?

Likewise, for scaling with a 0 factor.

joeng03: While it is possible to support negative scaling by patching the JSCAD code, Prof has said that he does not support the inclusion of such a functionality.

"I don't support the inclusion of negative scaling. It is not often used in 3D modelling. Would rather have a separate flip function to do that. Since all the primitive shapes we provide have some symmetry such that flipping can be achieved with rotation, a separate flip function or negative scaling is not necessary."

RichDom2185: Okay sure, if that's the reason then I guess that's fine. Thanks for the clarification!

joeng03: Hi @RichDom2185, when trying to merge this with the master branch, I get a syntax error when importing the CSG module in the frontend. I suspect this has something to do with the @jscad/modeling version, however, I have not found a way to resolve the merge conflict without causing this error. Could you please look into this? Thanks!

RichDom2185: What exactly is the command you are using to merge `master` into this branch? Your force pushes seem to be causing weird things in the GitHub PR list of changes.

RichDom2185: I will try to troubleshoot. Essentially, the module is done right? All that's left is to merge `master`?

joeng03: Yep, all that's left is to merge `master`

RichDom2185: Can you create a new branch with the contents of this PR just before your most recent force-pushed? (This one below)

image


What was the rationale of the force-push btw?

RichDom2185: Nevermind, was able to dig the commit out of GitHub's servers. See #213.

RichDom2185:
Superseded by #213.

joeng03:
Hi @RichDom2185 , thanks for solving the issues of this PR! 
As I am not that experienced in Git yet, I might have messed up the commit history by force-pushing. Could you explain what caused the syntax error issue after merging with the master branch, and how did you resolve it? Thanks!

RichDom2185:
> Hi @RichDom2185 , thanks for solving the issues of this PR! As I am not that experienced in Git yet, I might have messed up the commit history by force-pushing. Could you explain what caused the syntax error issue after merging with the master branch, and how did you resolve it? Thanks!

Hi, refer to my comment here: https://github.com/source-academy/modules/pull/206#discussion_r1208925709. I honestly did not know what command you used, so I can't comment on that (yet). But what I did was:

1. go back to an earlier version of this branch (thank goodness for the PR, cause it kept track of the history of force-pushes)
2. found that earlier commit on GitHub
3. created a new branch from that earlier commit
4. did an interactive rebase of that newly created branch on top of the latest `master`, ignoring conflicts on auto-generated files (e.g. `yarn.lock`, since they can be easily recreated later)
5. cherry-picked all remaining commits from the latest version
6. re-ran `yarn install` to update the `yarn.lock` without worrying about merge conflicts
7. re-formatted all files to fix their lint issues
8. manually resolved some compile errors from incorrect merge conflict resolutions

Of course, this is harder than it sounds. In reality, the first 2 interactive rebase attempts were unsuccessful because of an incorrect ordering of commits (this is what happens when you try to rebase after you have merged some branches – generally, either use merge commits all the way or rebase all the way to make your life easier, don't use/mix both methods). The main issues were:

* One of you kept committing the `.idea` files from your editor (WebStorm?), this should not be there in the first place to begin with, not only did it cause merge conflicts, it messes up other people's editors as well as the config files kept getting overwritten
* One of you has a different linter/formatter setup, that does not follow/respect the rules set for this codebase. For example, the files were constantly reformatted to use double quotes <kbd>"</kbd> instead of single quotes <kbd>'</kbd> for strings. In your commits, your formatter kept changing so many lines (not just the quotes, but other rules such as indentation, line length, etc, were all overridden), that it resulted in numerous merge conflicts in every step. Please ensure that your linter is properly set up so that it respects the rules set for the codebase, and also ensure your files are properly formatted before committing them. You should not wait until the very last commit to run `yarn lint --fix` to reformat all your files at once.

---

The above comment was mainly about the merge conflicts that I encountered while rebasing, and focused on how we can prevent conflicts (by configuring our editors properly).

In terms of how I resolved it, it is no other than just a simple rebase on top of `master` (essentially merging master into the branch), which is why I suspect you probably used the command wrongly, hence the above linked comment.

RichDom2185:
Based on the commit titles and hashes of this (and my experience debugging PRs), I strongly suspect somewhere along the way, the following happened:

* you tried to merge with `master`, but
* encountered merge conflicts, so
* you wanted to cancel the merge, and so
* undid all the merge changes (cleaning up your working directory), but
* did not do `git merge --abort` for some reason, and hence
* subsequently, when you wanted to commit a new change (bugfix/etc.),
* your next commit, which is supposed to be a feature commit,
* is created as a merge commit by git instead (because git still thinks you are merging), thus
* as a result, git thinks that
  * you have merged with `master` at that point in time,
  * but decided the right course of action was to ignore all the new changes from master,
  * and simply keep your changes
* thus the next time you tried to pull master again,
* you think you are `N` commits behind master,
* but git thinks you are only `N-some number` commits behind master (or even up to date), and
* if you try to rebase now, not only will that not help (because git thinks you are up to date), but
* you will also lose some of your actual, feature/fix/changes, because rebasing removes merge commits, and because you did not abort the merge earlier, git mistakenly marked what should be your normal commit as a merge commit

PR #210, opened 2023-05-09
Standardize line encodings

### LINKED ISSUE #209: Standardize Line Encodings
Moved from #204 

From @Cloud7050,

> Using git ls-files --eol, this repo appears to sometimes use LF and sometimes use CRLF in the index (and some files are even mixed). For comparison, the main js-slang repo appears to mainly use LF.
> 
> This repo's base eslint config is based on a config I've been carrying with me for a while. As I'm on Windows, that rule's setting is less of a concious choice for this repo. If I'm understanding this correctly, maybe it'll be better to:
> 
> - Rely on each user's git core.autocrlf config to handle line endings across platforms (true on Windows, input on Linux/macOS?)
> - Turn off the linebreak-style rule in the base eslint config. Users are free to locally checkout LF or CRLF
> - Push a commit that converts all existing files to be indexed with LF (could use something like [eol-converter-cli](https://www.npmjs.com/package/eol-converter-cli) on npm)
> From there I think the repo may sometimes get CRLF introduced, but they would sometimes get corrected to LF down the line. Any thoughts?

My perspective on this is informed by [Prettier](https://prettier.io/docs/en/options.html#end-of-line), which says:
> When people collaborate on a project from different operating systems, it becomes easy to end up with mixed line endings in a shared git repository. It is also possible for Windows users to accidentally change line endings in a previously committed file from LF to CRLF. Doing so produces a large git diff and thus makes the line-by-line history for a file (git blame) harder to explore.

Because of the above reason, I don't really agree with allowing for mixed line endings.

IMO, the best way to fix this issue will be to run prettier or eslint --fix in a pre-commit hook and configuring [.gitattributes](https://git-scm.com/docs/gitattributes#_text) to ignore line endings project-wide, and then we don't even need to push a commit to convert all existing files. 

Any thoughts?

Yongbeom-Kim:
# Description

Adds a `.gitattributes` file to standardize all line encoding to CRLF for the following files (LMK if there are any other important files to standardize, or any to omit):
- *.js
- *.ts
- *.tsx
- *.md
- *.cjs
- *.json

Fixes #209 
Sequenced after #204 (oops, sorry about that)

## Type of change

# How Has This Been Tested?

N/A

# Checklist:


RichDom2185:
> Sequenced after #204 (oops, sorry about that)

No worries, looks like there are now merge conflicts due to the squashing.

Yongbeom-Kim:
I've fixed the conflicts

From what I can see, there is no particular reason for CRLF, other than the fact that the current eslint rules specify CRLF in particular

.gitattributes
```diff
@@ -0,0 +1,12 @@
...
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Declare files that will always have CRLF line endings on checkout.
+*.js text eof=crlf
+*.ts text eof=crlf
+*.tsx text eof=crlf
+*.md text eof=crlf
+*.cjs text eof=crlf
+*.json text eof=crlf
```
shenyih0ng, line 12: i think there's a typo here:

```diff
+ eol=crlf
- eof=crlf
```

shenyih0ng (changes requested):

shenyih0ng:
if we want to take things even further we can introduce a [`.editorconfig` file](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/#bonus-create-an-editorconfig-file) to ensure new files always have the right line endings

ref (it's a great read): https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/

leeyi45:
> if we want to take things even further we can introduce a [`.editorconfig` file](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/#bonus-create-an-editorconfig-file) to ensure new files always have the right line endings
> 
> ref (it's a great read): https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/

Let us just go forward with adding this. I've been searching for the longest time for some way to get vscode to standardize line endings without having to commit `.vscode`. Good to see such a solution exists.

RichDom2185:
Let's recreate this branch directly off the latest `master`? Seems like we can just directly branch off `master` and cherry-pick ec377809a8f3ea3ffa6b108ce3e150c6bb584b69 instead of going through the hassle of dealing with all the merge conflicts from the recently-merged PRs.

martin-henz:
@Yongbeom-Kim what's the status of this PR?

martin-henz:
@Yongbeom-Kim what's the status of this PR?

Cloud7050:
@martin-henz to my understanding, @leeyi45 is incorporating these changes into #244. This original PR can probably be closed.

martin-henz:
> @martin-henz to my understanding, @leeyi45 is incorporating these changes into #244. This original PR can probably be closed.

OK

PR #215, opened 2023-06-22
Fix documentation

RichDom2185:
# Description

> We support markdown for documentation, so we should write our documentation in markdown, not HTML.

Fixes formatting of documentation (especially in Unity Academy). Fixes style inconsistencies as well as rendering bugs that are as a result of the HTML being rendered as-is.

This also makes the documentation compatible with our upcoming documentation repository,

## Type of change

### DIFF
--- a/src/bundles/unity_academy/index.ts
+++ b/src/bundles/unity_academy/index.ts
@@ -1,15 +1,14 @@
 /**
  * <b>A module that allows students to program with Unity Engine in either <u>2-D</u> or <u>3-D</u> scene in Source Academy.</b>
- * <br>
+ *
  * Makes use of "<b>Unity Academy Embedded WebGL Frontend</b>" (also made by me) to work together.<br>
- * <br>
+ *
  * <b>Code Examples: </b><a href = 'https://unity-academy.s3.ap-southeast-1.amazonaws.com/code_examples.html' rel="noopener noreferrer" target="_blank">Click Here</a><br>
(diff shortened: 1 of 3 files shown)


src/bundles/unity_academy/index.ts
```diff
@@ -1,32 +1,38 @@
...
+ * **Code Examples:** <a href = 'https://unity-academy.s3.ap-southeast-1.amazonaws.com/code_examples.html' rel="noopener noreferrer" target="_blank">Click Here</a><br>
+ * **Prefab Information:** <a href = 'https://unity-academy.s3.ap-southeast-1.amazonaws.com/webgl_assetbundles/prefab_info.html' rel="noopener noreferrer" target="_blank">Click Here</a><br>
+ * **User Agreement:** <a href = 'https://unity-academy.s3.ap-southeast-1.amazonaws.com/user_agreement.html' rel="noopener noreferrer" target="_blank">Click Here</a><br>
+ *
+ * **<u>Note that you need to use this module with a '<i>Native</i>' variant of Source language, otherwise you may get strange errors.</u>**
+ *
+ * **Lifecycle Event Functions**
+ *
+ * - Unity Academy has its own internal loop on students' GameObject lifecycle.
+ * - <u>Lifecycle event functions</u> are functions that are not supposed to be called by Source Academy's default evaluator, instead they are called by Unity Academy at certain time in the GameObject's lifecycle.
```
chownces, line 15: <u> tag still present

chownces (approved):

src/bundles/unity_academy/index.ts
RichDom2185: Should be fine as there's no Markdown alternative to `<u>`.

PR #220, opened 2023-09-04
Force game functions var args

gok99:
# Description

Module function var-args has been broken for awhile. Let's add a quick fix by injecting game functions `minArgsNeeded`

## Type of change

# How Has This Been Tested?

Load modules and check that var args are allowed

# Checklist:

### DIFF
--- a/src/bundles/game/functions.ts
+++ b/src/bundles/game/functions.ts
@@ -761,6 +761,7 @@ export default function gameFuncs(): { [name: string]: any } {
   Object.entries(functions)
     .forEach(([key, fn]) => {
       finalFunctions[key] = !scene ? nullFn : fn;
+      finalFunctions[key].minArgsNeeded = fn.length;
     });
 
   return finalFunctions;


lhw-1 (commented):
In conjunction with [this PR](https://github.com/source-academy/frontend/pull/2664), the changes seem to have fixed the erroneous behavior for "my room". Thanks for the PR!

Will leave it to the modules team to approve & merge.

leeyi45:
Before we go ahead with this, I do think it is worth looking into why the varargs problem exists, and what we can do to fix it for all modules without having to monkey patch things.

gok99:
I do agree - we'll only need this by week 7, so there is some time to do something more comprehensive. 

As far as I can tell, there seems to be no good way to get information about default parameters from a Function object. So, the arity checks in js-slang will require exactly the number of required parameters to be provided:

https://github.com/source-academy/js-slang/blob/2da5bd620e4af28c6dec516335a93c69bf424ad6/src/utils/operators.ts#L108-L118

For some context, I think module varargs used to work properly (?) long ago (probably before checks of this kind were introduced with rest/spread). When things broke, it started requiring all arguments (including those with default values) - I don't have much insight as to why. It now seems like that behavior has changed and exactly the required arguments must be provided.

martin-henz (approved):
OK, as discussed in telegram group "SA 2425 Game".

PR #239, opened 2023-09-10
sound and stereo_sound: renaming and fixing

martin-henz:
renaming play and play_concurrently to play_in_tab and play; fixing stereo_sound's play function; fixing user-facing comments

### DIFF
--- a/src/bundles/sound/functions.ts
+++ b/src/bundles/sound/functions.ts
@@ -426,8 +426,8 @@ export function play(sound: Sound): void {
     throw new Error(
       `${play.name} is expecting sound, but encountered ${sound}`,
     );
-  } else if (get_duration(sound) <= 0) {
-    // Do nothing
+  } else if (get_duration(sound) < 0) {
+    throw new Error(`${play.name}: duration of sound is negative`);
   } else {
     // Instantiate audio context if it has not been instantiated.
     if (!audioplayer) {
--- a/src/bundles/sound/functions.ts
+++ b/src/bundles/sound/functions.ts
@@ -1,19 +1,19 @@
 /**
- * The sounds library provides functions for constructing and playing sounds.
+ * The `sound` module provides functions for constructing and playing sounds.
  *
  * A wave is a function that takes in a number `t` and returns
  * a number representing the amplitude at time `t`.
  * The amplitude should fall within the range of [-1, 1].
  *
- * A Sound is a pair(wave, duration) where duration is the length of the sound in seconds.
+ * A Sound is a pair(wave, duration) where duration is the length of the Sound in seconds.
  * The constructor make_sound and accessors get_wave and get_duration are provided.
  *
  * Sound Discipline:
- * For all sounds, the wave function applied to and time `t` beyond its duration returns 0, that is:
+ * For all Sounds, the wave function applied to and time `t` beyond its duration returns 0, that is:
  * `(get_wave(sound))(get_duration(sound) + x) === 0` for any x >= 0.
  *
  * Two functions which combine Sounds, `consecutively` and `simultaneously` are given.
- * Additionally, we provide sound transformation functions `adsr` and `phase_mod`
+ * Additionally, we provide Sound transformation functions `adsr` and `phase_mod`
  * which take in a Sound and return a Sound.
  *
  * Finally, the provided `play` function takes in a Sound and plays it using your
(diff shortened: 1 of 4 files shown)


src/bundles/stereo_sound/functions.ts
```diff
@@ -252,7 +254,7 @@ export function record_for(duration: number, buffer: number): () => Sound {
  * @param right_wave wave function of the right channel of the sound
  * @param duration duration of the sound
  * @return resulting stereo sound
```
sayomaki, line 256: Maybe the return text should use `Sound` instead of `sound`?

src/bundles/stereo_sound/functions.ts
```diff
@@ -277,7 +279,7 @@ export function make_stereo_sound(
  * @param wave wave function of the sound
  * @param duration duration of the sound
  * @return with wave as wave function and duration as duration
```
sayomaki, line 281: The return documentation does not seem correct here.

src/bundles/stereo_sound/functions.ts
```diff
@@ -360,27 +360,28 @@ export function play_waves(
...
 }
 
 /**
  * Plays the given Sound using the computer’s sound device.
- * The sound is only played if no other sounds are currently being played.
+ * The sound is added to a list of sounds to be played one-at-a-time
+ * in a Source Academy tab.
  *
  * @param sound the sound to play
  * @return the given sound
```
sayomaki, line 373: `sound` here should be `Sound`.

src/bundles/stereo_sound/functions.ts
```diff
@@ -450,26 +451,13 @@ export function play(sound: Sound): AudioPlayed {
     riffwave.header.bitsPerSample = 16;
     riffwave.Make(channel);
 
-    /*
```
sayomaki, line 453: This part is still found it `sound/functions.ts` in L395, is it still needed there?

sayomaki (changes requested):
Being slightly nit-picky here, but I think the examples documentation changes are worth addressing.

martin-henz:
> Being slightly nit-picky here, but I think the examples documentation changes are worth addressing.

Thanks for the review @sayomaki . I think I've handled all comments.

src/bundles/sound/functions.ts
```diff
@@ -391,43 +392,31 @@ export function play(sound: Sound): AudioPlayed {
...
-      isPlaying = false;
-    }; */
-
     const soundToPlay = {
       toReplString: () => '<AudioPlayed>',
       dataUri: riffwave.dataURI,
     };
     audioPlayed.push(soundToPlay);
-    return soundToPlay;
+    return sound;
```
RichDom2185, line 400: Why was this REPL representation changed?

src/bundles/stereo_sound/functions.ts
```diff
@@ -450,96 +451,89 @@ export function play(sound: Sound): AudioPlayed {
...
-    }; */
-
     const audio = {
       toReplString: () => '<AudioPlayed>',
       dataUri: riffwave.dataURI,
     };
 
     audioPlayed.push(audio);
-    return audio;
+    return sound;
```
RichDom2185, line 460: Ditto regarding REPL representation.

RichDom2185 (commented):
Just some questions, thanks!

src/bundles/stereo_sound/functions.ts
```diff
@@ -450,96 +451,89 @@ export function play(sound: Sound): AudioPlayed {
...
+  } else if (get_duration(sound) < 0) {
+    throw new Error(`${play.name}: duration of sound is negative`);
   } else {
     // Instantiate audio context if it has not been instantiated.
     if (!audioplayer) {
       init_audioCtx();
     }
 
-    const channel: number[] = Array[2 * Math.ceil(FS * get_duration(sound))];
+    const channel: number[] = [];
```
sayomaki, line 487: There might be a slight issue here when accessing the `channel` array later on with positional accessors, as the array has no length. Maybe define the `len` first, then create the `channel` array with the `Array` constructor with `len`.

sayomaki (commented):
~~Just one more thing to address here, missed it out earlier. The rest looks good, thanks!~~

LGTM, thanks!

src/bundles/sound/functions.ts
martin-henz: Our "display" functions in modules are usually the identity function. I see the `play` functions to be in that category. The advantage is that you can debug your sounds by wrapping an expression that evaluates to a sound with a `play` call and it still will work.

src/bundles/stereo_sound/functions.ts
martin-henz: Ditto.

src/bundles/stereo_sound/functions.ts
sayomaki: On second thought, it may not be needed as the array will just leave empty items before the index.

sayomaki (approved):
LGTM, thanks!

PR #243, opened 2023-09-12
bug fix: zero duration sound

martin-henz:
Fix regression introduced in https://github.com/source-academy/modules/pull/239

Sounds with duration 0 caused error. Now working:

```
import { play, sine_sound } from 'sound';
play(sine_sound(400, 0));
```
and
```
import { play, sine_sound } from 'stereo_sound';
play(sine_sound(400, 0));
```


RichDom2185 (approved):
LGTM

PR #244, opened 2023-09-12
Miscellaneous Repo Improvements

### LINKED ISSUE #221: Generated documentation missing each module's summary
Modules user documentation used to include a general description of every module in that module's section of the documentation. These summaries come from the JSDoc comment at the top of the `functions.ts` file of each module bundle.

A recent change to the way documentation is generated has caused these summaries to no longer appear. Lee Yi has identified the possible point of regression as #153.

It is worth fixing as these summaries can be very informative and helpful. For example, curve's summary contains useful info like the definitions for curve, curve transformation, curve drawer, code snippets etc.

### LINKED ISSUE #225: [CURVE]: Calling animate with invalid CurveAnimation crashes frontend
A Source program that encounters errors when run will normally print a red error message to the REPL with the line number and relevant details such that the user can fix their code accordingly.

The curve module's animation functions however have the potential to crash the frontend. `animate_curve()` and `animate_3D_curve()` take in a `CurveAnimation` as their last argument, to provide them with curves over time. If there is an error while attempting to run the `CurveAnimation` callback, the frontend crashes.

Snippets to reproduce different-looking crashes:

```js
import { animate_curve, draw_connected } from 'curve';

animate_curve(1, 30, draw_connected(20), s => t => "invalid");

// A similar invalid snippet without the animation does not crash:
// draw_connected(20)(t => "invalid");
```

```js
import { animate_3D_curve, draw_3D_connected, make_3D_point } from 'curve';

// Notice how only 2 arguments are passed to make_3D_point()
animate_3D_curve(1, 30, draw_3D_connected(20), s => t => make_3D_point(1, 2));
```

image

image

leeyi45:
# External Packages
There are several packages available on the frontend and in js-slang, making it unnecessary for Source modules to include them when bundled. This PR will further externalize several of these packages (the full list [here](https://github.com/source-academy/modules/wiki/Build-System:-Bundles-and-Tabs)), reducing the size of files.

In the future, if there are packages that Source modules that require that are also already available at runtime, these packages should also be marked as external, but such changes would also require corresponding updates to the frontenda and/or js-slang

# Refinement of Module Scripts
The script system has become increasingly complex. To aid with this I've condensed all the script management code into a single Javascript file. Instead of committing the entire output of compiling module scripts, module scripts can instead be built using the `yarn scripts:build` command (which should be run automatically on `postinstall`). Module script functionality remains identical.

`yarn test` and `yarn scripts:test` should now both work with Windows paths, where previously Jest only recognized posix style paths as valid patterns.

# Error Handling for `AnimationCanvas`
Animations are calculated at runtime, so there is no way for bundle code to detect beforehand if an animation is invalid or will throw an error. As such, an error screen has been added to the various animation canvases which will allow these to throw errors gracefully. Fixes #225.

image

# Change in `throwIfNotRune`
Similar to how the functions in `sound` have been modified to allow for custom function names, the `throwIfNotRune` function in the `rune` bundle has also been modified to allow custom function names to be shown when errors are thrown.

# Updated `typedoc`
Typedoc has since fixed the issue where the README was not getting generated with the rest of the HTML documentation. The original README file has thus been re-included.

# Documentation Relocation
The summaries that used to be present in the `functions.ts` files of bundles have been relocated to their entry points. These summaries should now be visible within the HTML documentation. Resolves #221.

# Tab Development Server
Currently, the frontend is the only way to render module tabs. This means that every change requires a rebuild of the tab, followed by a refresh on the frontend.

This PR will introduce a Vite server that is able to react to changes in tabs and render them live.

For the server to be able to work, components need to be defined as a separate constant, instead of just within the `body()` function:
```tsx
export default {
  body; () => <div></div>
}
```
should be defined like this:
```tsx
const Component = () => <div></div>
export default {
  body; () => <Component />
}
```
Use `yarn devserver` to start the development server.
# Standardize Line Encoding
This PR also incorporates the changes from #210. With the addition of `.editorconfig` and `.gitattributes` hopefully development across OSes should be more consistent.


leeyi45:
Does anyone happen to know what to do about running Jest snapshots over the CI

leeyi45:
Figured out the issue with the CI: On Windows file paths are case-insensitive and the snapshot file was named `curve.tsx.snap`. When running on Windows this is not an issue, but this doesn't work with posix systems, so it was failing the CI. I finally noticed the discrepancy.

src/bundles/rune/runes_ops.ts
```diff
@@ -1,361 +1,361 @@
...
+ * primitive Rune in the rune of a heart
+ * */
+export const getHeart: () => Rune = () => {
+  const vertexList: number[] = [];
+  const colorList: number[] = [];
+
+  const root2 = Math.sqrt(2);
+  const r = 4 / (2 + 3 * root2);
+  const scaleX = 1 / (r * (1 + root2 / 2));
+  const numPoints = 10;
```
RichDom2185, line 197: Was this reversion intentional?

RichDom2185 (approved):
LGTM, just one minor comment as below. Could you take a look at the changes I pushed as well, and if it's all good to you, feel free to merge.

Thanks for working on this!

src/bundles/rune/runes_ops.ts
leeyi45: This was part of the change to increase rune resolution? My Windows machine has been having some issues with the line endings, so the merge conflict may have been resolved incorrectly.

RichDom2185 (approved):
LGTM, thanks!

PR #234, opened 2023-09-06
Revamp CSG functions, make curve canvases consistent & more

Cloud7050:
The changes in this PR are described below.

### CSG

Changes in: `package.json`, `src/bundles/csg/`, `src/common/utilities.ts`, `src/tabs/Csg/`, `yarn.lock`.

- Revamp CSG exports
  - Add new ability to `ungroup()` Groups while applying accumulated transforms
  - Change centring of some primitives for better usability and consistency, instead of aligning all min bounds to origin, as not all primitives have full 1x1x1 bounding box dimensions. Closes Cloud7050/modules#30
    - Geodesic sphere & star now centred at (0.5, 0.5, 0.5) instead of aligning to origin
    - Square pyramid now has base length 1 with sides aligned to axes. No longer rotated 45° with corners touching 1x1 box by way of low resolution cone generator
    - Prism now centred instead of aligning to origin. Flat side now aligned with y axis instead of x
    - Torus now centred higher at (0.5, 0.5, 0.5). Ring is now slightly thicker, while maintaining 1x1 dimensions
  - Rewrite all user documentation. Fix various outdated/wrong information. Add significantly more detail. Closes Cloud7050/modules#33
    - Add proper summary to documentation, with sample usage and link to all GitHub samples. Closes #230
    - Fix primitives losing `@category`. Add fitting categories to all exports
  - Officially support rendering directly passed Shapes as opposed to only Groups
  - Add more parameter checks to avoid likely user pitfalls and reduce unhelpful error messages. Enhance existing error messages and increase specificity
  - Rename `shape_to_stl()` to `download_shape_stl()` as it triggers a file download
  - Rename V2 `Entity` to `Operable` to avoid ambiguity with renderer and JSCAD internal entities. Operables are now a user-facing term
- Finally upgrade to real BlueprintJS tooltips now that we have the technology. Fixes buggy tooltip positions that were meant to be a temporary measure. Tweak control hint text. Closes Cloud7050/modules#22
- Upgrade to real `js-slang` List types and related functions now that we have the technology
- Split samples into their own files. Add many new samples. Update existing samples to improve them and make them work with current CSG. Closes #231
- Refactor, simplify, and move implementations, such as for primitives and utilities
  - Lists now only used in main functions file, as importing a file that works with real Lists from the tab end causes build failures. Add note regarding limitation. Groups now store arrays instead of Lists
  - Remove Entity cloning
  - Remove unused code
- Move relevant CSS constants into their own common file. Update them for BlueprintJS V4
[...]

PR #250, opened 2023-10-01
Revert runtime-error regression in zero duration

kjw142857:
# Description

Revert regression that caused make_sound to reject durations of length 0, which affected the consecutively and simultaneously functions

Note: Issue was caused by #234

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)


RichDom2185:
Fixes the regression introduced in #234.

RichDom2185 (approved):
LGTM, thanks!

PR #252, opened 2023-10-08
scrabble: include all words

martin-henz:
updating the scrabble module to contain an array of all scrabble words. Previously only words of size 10 and up were included. Removed list of words for performance. So we only include an array of words: will be useful for a binary search assessment.


src/bundles/scrabble/__tests__/index.ts
```diff
@@ -1,17 +1,9 @@
-import { scrabble_array, scrabble_list } from '../functions';
-
-function list_ref(list, n) {
-  return n === 0 ? list[0] : list_ref(list[1], n - 1);
-}
+import { scrabble_words } from '../functions';
 
 // Test functions
 
 test('get the first word in the array', () => {
```
RichDom2185, line 5: Perhaps we can change the test description

RichDom2185 (approved):
LGTM, with just one minor comment:

PR #262, opened 2023-10-31
Fix the bug that AceEditor in REPL tab is not notified during resize

larrywang0701:
Fix the bug that the actual code area in AceEditor in REPL tab not resizes with the displayed editor area. Especially when extending the editor to a larger height, the new display area added compared to initial area will all become invalid code area that cannot display code.
image
Now AceEditor will automatically resize the code area when user drag the resize bar. When extend the editor to a larger height, all area belongs to the editor can normally display code.
image


martin-henz (approved):
looks good!

PR #263, opened 2023-10-31
Bug fix and changes to REPL module

larrywang0701:
- Removed "Program exited with return value" message as Professor Martin instructed
 - Fix a bug that causes a runtime error to be shown in the browser's console when student provide an evaluator entry function with incorrect parameter count
   - For example: `set_evaluator(()=>undefined);` or `set_evaluator((param1,param2)=>undefined);`
   - Now the REPL will display the error message `Error: Unable to use your evaluator to run the code. Does your evaluator entry function contains and only contains exactly one parameter?` to students if they make this mistake in their code.
- Other minor changes:
   - Removed "metacircular" in documentation since the module is not only limited to metacircular evaluator, but also possibly on CSE machine and any evaluators that use a function with one parameter to accept raw code as entry.
   - In the error message: the function name `invoke_repl` should be `set_evaluator`

### DIFF
--- a/src/bundles/repl/functions.ts
+++ b/src/bundles/repl/functions.ts
@@ -11,8 +11,15 @@ import { COLOR_REPL_DISPLAY_DEFAULT } from './config';
 const INSTANCE = new ProgrammableRepl();
 context.moduleContexts.repl.state = INSTANCE;
 /**
- * Setup the programmable REPL with given metacircular evaulator entrance function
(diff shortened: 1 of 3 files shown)


martin-henz (approved):
All looking good! Nice job @larrywang0701 
The REPL module will be very useful in future versions of CS1101S and CS4215.

PR #266, opened 2023-11-19
Vector calculus

srj31:


martin-henz:
@srj31 confirmed that this draft is not needed any longer.

PR #271, opened 2024-01-26
Update Repository to Node 20

leeyi45:
Bumping packages and runtimes to Node 20. If this breaks anybody's modules please sound out, I personally have no way to test the functionalities of each and every module on my own.


RichDom2185 (approved):
LGTM, thanks! I think we can just merge this so that we can get everyone to test in their local dev server; CS1101S uses a frozen copy so it's not going to break unless we manually deploy.

RichDom2185:
@leeyi45 update: it doesn't work (see below):

image

I'll create a PR to revert this, oops.

PR #276, opened 2024-02-18
Refactor scrabble module

RichDom2185:
# Description

Generates scrabble letters using `.map` on words to:
* Improve editor responsiveness (smaller tokenization, less number of lines, etc.)
* Allow for single source of truth

Also generates the tiny dataset using `.filter` every 100 words.

The correctness is guaranteed by the introduction of additional tests, ensuring they stay the (snapshot-checking).

## Type of change


RichDom2185 (commented):
I've generated the snapshots once before the refactor (refer to commit history) and can confirm that the refactor does not change the snapshots.

RichDom2185:
@martin-henz can I check what exactly qualifies for `scrabble_words_tiny`? Or is it handpicked? Perhaps we can generate it in a similar way, using array functions on `scrabble_words`.

martin-henz:
> @martin-henz can I check what exactly qualifies for `scrabble_words_tiny`? Or is it handpicked? Perhaps we can generate it in a similar way, using array functions on `scrabble_words`.

I remember playing with the scrabble module in preparation for the PA last semester. There was a concern that the full dataset would be too large for some student's computers, so I came up with scrabble_words_tiny. I think I sampled the full set at intervals of 100:
```
display(array_length(scrabble_words));          // 172820
display(array_length(scrabble_words_tiny));     // 1729
```

RichDom2185:
> > @martin-henz can I check what exactly qualifies for `scrabble_words_tiny`? Or is it handpicked? Perhaps we can generate it in a similar way, using array functions on `scrabble_words`.
> 
> I remember playing with the scrabble module in preparation for the PA last semester. There was a concern that the full dataset would be too large for some student's computers, so I came up with scrabble_words_tiny. I think I sampled the full set at intervals of 100:
> 
> ```
> display(array_length(scrabble_words));          // 172820
> display(array_length(scrabble_words_tiny));     // 1729
> ```
> 
> [We ended up not using the `scrabble` module for the PA.]

Ok, I've used `.filter((_, i) => i % 100 === 0)` to generate `scrabble_words_tiny` and it seems to be an exact match with the snapshot.

This PR is ready now, thank you!

martin-henz (approved):
Nice!

PR #281, opened 2024-02-22
🐛 fix: incorrect rgb colors

### LINKED ISSUE #279: Plotly: Colors not showing
# Prerequisites

Please answer the following questions for yourself before submitting an issue. **YOU MAY DELETE THE PREREQUISITES SECTION.**

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

# Expected Behavior

See https://share.sourceacademy.org/2p4j8

# Current Behavior

Colors not showing when using plotly:draw_points_3d

# Failure Information (for bugs)

Screenshot 2024-02-22 at 10 34 04 AM

srj31:
Plotly treats decimal numbers for rgb value weirdly, where if rgb values are (0.9, 0, 0) it uses 0.9 as a percentage to display bright red

Fixes #279 
## Type of change

### DIFF
--- a/src/bundles/plotly/curve_functions.ts
+++ b/src/bundles/plotly/curve_functions.ts
@@ -47,7 +47,7 @@ export function z_of(pt: Point): number {
  * ```
  */
 export function r_of(pt: Point): number {
-  return pt.color[0] ?? 0 * 255;
+  return ( pt.color[0] ?? 0 ) * 255;
 }
 
 /**
@@ -62,7 +62,7 @@ export function r_of(pt: Point): number {
  * ```
  */
 export function g_of(pt: Point): number {
-  return pt.color[1] ?? 0 * 255;
+  return ( pt.color[1] ?? 0 ) * 255;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


RichDom2185 (commented):
The change does need seem correspondent with the PR description; how does the fix for operation ordering solve the issue?

martin-henz:
Well, the description is a bit strange. It's clear that the original implemention intended to scale the color values by a factor of 255. The original implementation assumed that ?? has higher precedence than *. However, ?? has lower precedence than *. So 
```
pt.color[1] ?? 0 * 255;
```
is parsed like
```
pt.color[1] ?? (0 * 255);
````
which was obviously wrong because the color value isn't scaled at all. The PR fixes this bug.

RichDom2185:
> Well, the description is a bit strange.

Yeah, it is. But I agree, the operation ordering is definitely incorrect. I guess we can merge this first and reopen the issue if it still persists.

RichDom2185 (approved):

srj31:
I wanted to explain why the error was not caught before when tested with different inputs. For e.g. `pt.color[0] ?? 0 * 255` would return `0.9` which plotly would convert it to the `0...255` range value and it looked as if it worked correctly.

martin-henz:
Ah, makes sense. Thanks!

martin-henz:
> I wanted to explain why the error was not caught before when tested with different inputs. For e.g. `pt.color[0] ?? 0 * 255` would return `0.9` which plotly would convert it to the `0...255` range value and it looked as if it worked correctly.

It's fixed.

PR #283, opened 2024-03-01
Added communication library

8kdesign:
# Description

Added communication module.

The communication module allows for real-time interaction between different users, creating new possibilities for what users can do on Source Academy. 

## Type of change

# How Has This Been Tested?

- Added testing for communications module logic
- Created sample scenarios for usage of modules in Source Academy

# Checklist:


src/bundles/communication/Communications.ts
```diff
@@ -0,0 +1,123 @@
...
+  rpc: RpcController | null = null;
+
+  constructor(address: string, port: number) {
+    let multiUser = new MultiUserController();
+    multiUser.setupController(address, port);
+    this.multiUser = multiUser;
+  }
+}
+
+let moduleState = new CommunicationModuleState('broker.hivemq.com', 8884);
```
RichDom2185, line 18: Can this be a `const` instead? Looking through the file, I don't see it being reassigned anywhere. Unless we plan to allow the user to specify which broker to use, in which case we should also allow users to specify the port number in addition to the host.

RichDom2185, line 18: I believe MQTT's default port is 1883 (and 9001 by convention for MQTT over WebSockets)? Why are we using 8884 here? Is it a special port for something? Would be good to add a comment for documentation.

package.json
```diff
@@ -131,6 +136,7 @@
   },
   "resolutions": {
     "@types/react": "^18.2.0",
-    "esbuild": "^0.18.20"
+    "esbuild": "^0.18.20",
+    "**/gl": "^6.0.2"
```
RichDom2185, line 140: Modules is still using Node 16 at the moment, so I don't think this GL dependency resolution is necessary.

src/bundles/communication/Communications.ts
```diff
@@ -0,0 +1,123 @@
...
+ */
+export function keepRunning() {
+  interval = window.setInterval(() => {}, 20000);
+}
+
+/**
+ * Removes interval that keeps the program running.
+ */
+export function stopRunning() {
+  if (interval) {
```
RichDom2185, line 36: Is the ID returned from `setInterval` always greater than 0? If not (or unsure), I think it would be best to explicitly state `interval !== undefined` as the condition.

src/bundles/communication/GlobalStateController.ts
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @param topicHeader Identifier for all global state messages, must not include '/'.
+ * @param multiUser Instance of multi user controller.
+ * @param callback Callback called when the global state changes.
+ */
+export class GlobalStateController {
+  private topicHeader: string;
+  private multiUser: MultiUserController;
+  private callback: (state: any) => void;
+  globalState: any;
```
RichDom2185, line 15: Is there any way to avoid `any`? Likewise, in the callback passed to the `initGlobalState` function in `Communications.ts`

src/bundles/communication/Communications.ts
```diff
@@ -0,0 +1,123 @@
...
+
+// Global State
+
+/**
+ * Initializes global state.
+ *
+ * @param topicHeader MQTT topic to use for global state.
+ * @param callback Callback to receive updates of global state.
+ */
+export function initGlobalState(
```
RichDom2185, line 50: Since this is to be exported, might be good to add a guard clause to prevent it being called a second time (perhaps throw an error to the user)?

src/bundles/communication/RpcController.ts
```diff
@@ -0,0 +1,123 @@
...
+export class RpcController {
+  private topicHeader: string;
+  private multiUser: MultiUserController;
+  private userId: string;
+  private functions = new Map<string, DeclaredFunction>();
+  private pendingReturns = new Map<string, (args: any[]) => void>();
+  private returnTopic: string;
+
+  constructor(
+    topicHeader: string,
```
RichDom2185, line 25: Are we going to allow arbitrary, user-defined MQTT topics? This might pose a problem with our own AWS MQTT broker, specifically, due to the whitelist. Though this shouldn't be an issue for other brokers, but just something to take note.

src/bundles/communication/GlobalStateController.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Sets up callback for global state messages.
+   * Parses received message and stores it as global state.
+   */
+  private setupGlobalState() {
+    if (this.topicHeader.length <= 0) return;
+    this.multiUser.addMessageCallback(this.topicHeader, (topic, message) => {
+      let shortenedTopic = topic.substring(
```
RichDom2185, line 35: Ditto on preferring `const` where possible.

src/bundles/communication/GlobalStateController.ts
```diff
@@ -0,0 +1,117 @@
...
+      let newGlobalState = { ...this.globalState };
+      if (
+        this.globalState instanceof Array ||
+        typeof this.globalState === 'string'
+      ) {
+        newGlobalState = {};
+      }
+      let currentJson = newGlobalState;
+      for (let i = 1; i < splitTopic.length - 1; i++) {
+        let subTopic = splitTopic[i];
```
RichDom2185, line 67: Ditto on `const` preference.

src/bundles/communication/GlobalStateController.ts
```diff
@@ -0,0 +1,117 @@
...
+          typeof currentJson[subTopic] === 'string'
+        ) {
+          currentJson[subTopic] = {};
+        }
+        currentJson = currentJson[subTopic];
+      }
+      if (message === undefined || message.length === 0) {
+        delete currentJson[splitTopic[splitTopic.length - 1]];
+      } else {
+        let jsonMessage = JSON.parse(message);
```
RichDom2185, line 80: Ditto.

src/bundles/communication/GlobalStateController.ts
```diff
@@ -0,0 +1,117 @@
...
+        currentJson[splitTopic[splitTopic.length - 1]] = jsonMessage;
+      }
+      this.setGlobalState(newGlobalState);
+    } catch {}
```
RichDom2185, line 84: Is the empty `catch` intentional?

src/bundles/communication/MqttController.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @param messageCallback Callback when a message has been received.
+ */
+export class MqttController {
+  private client: MqttClient | null = null;
+  private connected: boolean = false;
+  private connectionCallback: (status: string) => void;
+  private messageCallback: (topic: string, message: string) => void;
+
+  address: string = '';
+  port: number = 8080;
```
RichDom2185, line 21: MQTT default port should be 1883, unless there's a specific reason we're using 8080, which in that case, should be documented.

src/bundles/communication/MqttController.ts
```diff
@@ -0,0 +1,103 @@
...
+      this.connectionCallback(STATE_DISCONNECTED);
+    });
+    this.client.on('reconnect', () => {
+      this.connectionCallback(STATE_RECONNECTED);
+    });
+    this.client.on('offline', () => {
+      this.connectionCallback(STATE_OFFLINE);
+    });
+    this.client.on('message', (topic, message) => {
+      let decoder = new TextDecoder('utf-8');
```
RichDom2185, line 53: Ditto on `const`.

src/bundles/communication/MqttController.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.messageCallback = messageCallback;
+  }
+
+  /**
+   * Sets up MQTT client link and connects to it.
+   * Also handles connection status callbacks.
+   */
+  public connectClient() {
+    if (this.connected || this.address.length === 0) return;
+    let link = `wss://${this.address}:${this.port}/mqtt`;
```
RichDom2185, line 37: Ditto on `const`.

src/bundles/communication/RpcController.ts
```diff
@@ -0,0 +1,123 @@
...
+    userId?: string,
+  ) {
+    this.topicHeader = topicHeader;
+    this.multiUser = multiUser;
+    this.userId = userId ?? uniqid();
+    this.returnTopic = `${this.topicHeader}_return/${this.userId}`;
+    this.multiUser.addMessageCallback(this.returnTopic, (topic, message) => {
+      let messageJson = JSON.parse(message);
+      let callId = messageJson.callId;
+      let callback = this.pendingReturns.get(callId);
```
RichDom2185, line 36: Ditto on `const`.

src/bundles/communication/RpcController.ts
```diff
@@ -0,0 +1,123 @@
...
+   * @param sender ID of caller.
+   * @param callId ID of function call.
+   * @param result Return value for function call.
+   */
+  private returnResponse(sender: string, callId: string, result: any) {
+    let message = {
+      callId,
+      result,
+    };
+    let topic = `${this.topicHeader}_return/${sender}`;
```
RichDom2185, line 56: Ditto on `const`.

src/bundles/communication/MqttController.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (this.client) {
+      this.client.end(true);
+    }
+    this.connectionCallback = () => {};
+    this.messageCallback = () => {};
+  }
+
+  /**
+   * Broadcasts message to topic.
+   * QoS of 1, all listening devices receive the message at least once.
```
RichDom2185, line 71: Doesn't QoS 1 only guarantee that the sender sends at least once (and the broker receives at least once)? It's up to the QoS between the broker and the subscriber at the point of subscription to determine whether the QoS stays at 1 or gets downgraded between the broker and subscriber.

src/bundles/communication/MqttController.ts
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @param topic Identifier for group of devices to broadcast to.
+   * @param message Message to broadcast.
+   * @param isRetain Whether the message should be retained.
+   */
+  public publish(topic: string, message: string, isRetain: boolean) {
+    this.client?.publish(topic, message, {
```
RichDom2185, line 78: Can we guarantee that a connection is always successful at this point? Otherwise, due to QoS 1, the client might keep unnecessarily spamming the connection if it's unsuccessful (as it expects a PUBACK).

src/bundles/communication/MultiUserController.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Sets up and connect to the MQTT link.
+   * Uses websocket implementation.
+   *
+   * @param address Address to connect to.
+   * @param port MQTT port number.
+   */
+  public setupController(address: string, port: number) {
+    let currentController = this.controller;
```
RichDom2185, line 21: Ditto on `const`, and a couple more below – I won't mark them all.

src/bundles/communication/__tests__/index.ts
```diff
@@ -0,0 +1,224 @@
+import { MultiUserController } from '../MultiUserController';
+import { GlobalStateController } from '../GlobalStateController';
+
+let multiUser = new MultiUserController();
+multiUser.setupController('broker.hivemq.com', 8884);
```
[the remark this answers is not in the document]
RichDom2185: This is a test file, so I think it's fine to hardcode.

RichDom2185 (changes requested):
Hi, thanks for working on this! I have some comments below, let me know your thoughts!

src/bundles/communication/__tests__/index.ts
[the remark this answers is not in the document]
martin-henz: I think there are two distinct use-cases:

(1) playing with distributed programming in sourceacademy.org, using a public MQTT server.

(2) using our own MQTT server from sourceacademy.nus.edu.sg.

I think (1) is enough for @8kdesign 's project, so let's focus on that and get the PR merged. 

Fore (2): It would be great to re-use (1) somehow.


|

src/bundles/communication/MqttController.ts
```diff
@@ -0,0 +1,107 @@
+import { connect, type MqttClient } from 'mqtt/dist/mqtt';
```
leeyi45, line 1: Please add a comment explaining why this is used over just `mqtt`

8kdesign (commented):
Added comments

src/bundles/communication/Communications.ts
martin-henz: What @RichDom2185 means is that if `interval` can become 0 (also considered `false`) you should write `interval !== undefined`.

src/bundles/communication/GlobalStateController.ts
martin-henz: Yes, 'any' is quite justified here. The state can literally be **anything**.

src/bundles/communication/Communications.ts
martin-henz: ok

src/bundles/communication/RpcController.ts
martin-henz: Yes, we need to advise students accordingly. That's beyond this PR.

src/bundles/communication/Communications.ts
8kdesign: Ok fixed

martin-henz (approved):
All good.

PR #285, opened 2024-03-11
Fix Broken Patch for CSG After ESBuild Bump

### LINKED ISSUE #265: [CSG]: Recent changes broke patch's colorizePolygons
When attempting to run the [snowglobe sample](https://github.com/Cloud7050/modules/blob/949de19003be2b935e44727f11f9ee5913dd990c/src/bundles/csg/samples/snowglobe.js), the following error is shown in the REPL:

> Line 11: ReferenceError: colorizePolygons is not defined

That name occurs in the JSCAD modelling library's patch file. It would appear that recent changes in master may have broken the CSG module.

Line 11 calls `intersect()` from the CSG module. A trivial example such as `render(cube(silver))` still works without error.

joeng03:
# Description
This fixes 2 Issues:

1. The `@jscad/stl-serializer` package depends on `@jscad/modeling`. It will resolve the version of `@jscad/modeling` to 2.9.11. To reduce bundle size and standardize the version of `@jscad/modeling` at 2.9.6 for the patch, I downgraded the version of `@jscad/stl-serializer`.

2. The `esbuild` version bump seemed to have failed the function call to ` toPolygons(geometry1, colorizePolygons = true) `.  Passing in `colorizePolygons = true` is incorrect JavaScript syntax, initially written with named parameters in mind (but named parameters are not supported in JavaScript). It runs and produces the expected output without throwing a `ReferenceError` before the  `esbuild` version bump. I have rectified the syntax and created a new patch.

Fixes #265 

## Type of change

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "@box2d/debug-draw": "^0.10.0",
     "@jscad/modeling": "2.9.6",
     "@jscad/regl-renderer": "^2.6.1",
-    "@jscad/stl-serializer": "^2.1.13",
+    "@jscad/stl-serializer": "2.1.11",
     "ace-builds": "^1.25.1",
     "classnames": "^2.3.1",
     "dayjs": "^1.10.4",
--- a/patches/@jscad+modeling+2.9.6.patch
+++ b/patches/@jscad+modeling+2.9.6.patch
@@ -19,20 +19,15 @@ index 85720f4..9ef2a9d 100644
 -declare function toPolygons(geometry: Geom3): Array<Poly3>
 +declare function toPolygons(geometry: Geom3, colorizePolygons: boolean): Array<Poly3>
 diff --git a/node_modules/@jscad/modeling/src/geometries/geom3/toPolygons.js b/node_modules/@jscad/modeling/src/geometries/geom3/toPolygons.js
-index 25f9f50..300aaa4 100644
+index 25f9f50..bdb9f16 100644
 --- a/node_modules/@jscad/modeling/src/geometries/geom3/toPolygons.js
 +++ b/node_modules/@jscad/modeling/src/geometries/geom3/toPolygons.js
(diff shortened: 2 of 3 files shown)


RichDom2185 (commented):
Did you manually edit the lockfile? Can we just use a dependency resolution in package.json instead to ensure it doesn't accidentally get updated automatically?

joeng03:
> Did you manually edit the lockfile? Can we just use a dependency resolution in package.json instead to ensure it doesn't accidentally get updated automatically?

Which lockfile did you mean, the .patch file or yarn.lock? I didn't manually edit either one

RichDom2185:
> > Did you manually edit the lockfile? Can we just use a dependency resolution in package.json instead to ensure it doesn't accidentally get updated automatically?
> 
> Which lockfile did you mean, the .patch file or yarn.lock? I didn't manually edit either one

I mean yarn.lock.

joeng03:
Since the patch file applies the changes on fixed line numbers, it is hard to guarantee that future versions will not cause the patched package to break due to wrongly replacing newly added code.

RichDom2185:
> Since the patch file applies the changes on fixed line numbers, it is hard to guarantee that future versions will not cause the patched package to break due to wrongly replacing newly added code.

So we only need to freeze `@jscad/modeling`, right? Is there a reason to freeze `@jscad/stl-serializer`?

joeng03:
`@jscad/stl-serializer` depends on `@jscad/modeling` but of a different version (2.9.11), so it may be better to standardize the version at 2.9.6

RichDom2185 (approved):
> The latest version of`@jscad/stl-serializer` depends on `@jscad/modeling` but of a different version (2.9.11), so it may be better to standardize the version to 2.9.6

I was thinking about it again and you're right, it doesn't make sense to only freeze the transitive dependency. Thanks!

PR #286, opened 2024-03-11
Added AR library

8kdesign:
# Description

Added AR module.
The AR module allows users to run an instance of AR on their device (currently Android only, can run in simulator on desktop browser) and add objects to it.

## Type of change

# How Has This Been Tested?

- Created sample AR scenarios within Source Academy Playground

Note: Please make sure your frontend is running on HTTPS if testing locally. To run the frontend server on HTTPS, refer to [this](https://stackoverflow.com/questions/44574399/create-react-app-how-to-use-https-instead-of-http)

# Checklist:

- [ ] New and existing unit tests pass locally with my changes


src/bundles/ar/libraries/object_state_library/ARObject.tsx
```diff
@@ -0,0 +1,394 @@
...
+    newObject = UIObject.parseObject(object);
+    if (newObject) {
+      newObject.isSelected = isSelected;
+      return newObject;
+    }
+    newObject = LightObject.parseObject(object);
+    if (newObject) {
+      newObject.isSelected = isSelected;
+      return newObject;
+    }
```
joeng03, line 88: This could be simplified to 

```
    let newObject = CubeObject.parseObject(object) ||
                    SphereObject.parseObject(object) ||
                    GltfObject.parseObject(object) ||
                    UIObject.parseObject(object) ||
                    LightObject.parseObject(object);

    if (newObject) {
        newObject.isSelected = isSelected;
        return newObject;
    }
```

src/bundles/ar/AR.ts
```diff
@@ -0,0 +1,272 @@
...
+  if (!(json instanceof Object)) return;
+  const objects: ARObject[] = [];
+  const keys = Object.keys(json);
+  keys.forEach((key) => {
+    const item = json[key];
+    const parsedObject = ARObject.fromObject(item);
+    if (parsedObject) {
+      objects.push(parsedObject);
+    }
+  });
```
joeng03, line 188: This could be simplified using Object.entries(json)

src/bundles/ar/AR.ts
```diff
@@ -0,0 +1,272 @@
...
+    return; // Already in array
+  }
+  if (object.onSelect) {
+    moduleState.clickCallbacks.set(object.id, () => {
+      object.onSelect?.(object);
+    });
+  }
+  const newArray = Object.assign([], moduleState.arObjects);
+  newArray.push(object);
+  moduleState.arObjects = newArray;
```
joeng03, line 146: Does the `moduleState` need to be immutable? If there is no specific reason to keep it immutable under the hood, then we could actually make it mutable to improve performance (Tried rebuilding the module using mutable moduleState, works fine for me). 

Similar for other instances in AR.ts where the moduleState.arObjects is completely replaced, like `removeARObject()`

joeng03 (changes requested):
Solid PR! A few suggestions in terms of code quality and implementation: 



1. Multiple set/remove functions for toggle buttons (left/center/right). We could reduce it to 2 functions, each taking an extra string parameter("left" | "center" | "right").

2. Place all React components into tabs/

3. Abstract out components into their separate files, for example `ButtonComponent`

4. Change `object: ARObject` to ` arObject: ARObject` to avoid confusion in naming with the JS `object`

src/bundles/ar/AR.ts
8kdesign: The ARObject.fromObject static function here is custom defined. I don't think this world work

src/bundles/ar/AR.ts
8kdesign: The arObjects here can be modified while the AR context is running. (For example, one could add new objects to the world at the click of a button)

src/bundles/ar/AR.ts
joeng03: I meant we should use Object.values() to replace Object.keys() so the line 
  const item = json[key];
  could be removed. The rest of the code should stay the same.

src/bundles/ar/AR.ts
```diff
@@ -0,0 +1,297 @@
...
+
+/**
+ * Sets AR objects from json.
+ */
+export function setAsARObjects(json: any) {
+  const moduleState = getModuleState();
+  if (!moduleState) return;
+  if (!(json instanceof Object)) return;
+  const objects: ARObject[] = [];
+  for (const [_, item] of Object.entries(json)) {
```
joeng03, line 178: Use Object.values() to avoid creating an unused variable

joeng03 (commented):
LGTM overall. Just one nit to fix:

joeng03 (approved):

PR #290, opened 2024-03-13
[CSG] Add Empty Shape

### LINKED ISSUE #289: [CSG]: empty shape
I think we should have a constant `empty` that refers to a shape that consists of nothing.

Example usage:

https://share.sourceacademy.org/u6mq1

joeng03:
# Description

Fixes #289

Note: `yarn lint --fix` was run so some of the CSG files were also linted accordingly.

## Type of change

# How Has This Been Tested?

```
// Menger sponge, see
// https://en.wikipedia.org/wiki/Menger_sponge

import { union, subtract,intersect, translate, rotate, scale, render, 
         cube, blue,rounded_cube, empty_shape }
from 'csg';

const c = cube('#88ccff');

const d = scale(cube(blue), 1.5, 1.5, 1.5);

const empty = empty_shape();

function beside(s1, s2) {
    return union(scale(s1, 0.5, 1, 1),
                 translate(scale(s2, 0.5, 1, 1),
                           0.5, 0, 0));
}
function beside_frac(f, s1, s2) {
    return union(scale(s1, f, 1, 1),
                 translate(scale(s2, 1 - f, 1, 1),
                           f, 0, 0));
}
function before(s1, s2) {
    return union(scale(s1, 1, 0.5, 1),
                 translate(scale(s2, 1, 0.5, 1),
                           0, 0.5, 0));
}
function before_frac(f, s1, s2) {
    return union(scale(s1, 1, f, 1),
                 translate(scale(s2, 1, 1 - f, 1),
                           0, f, 0));
}

function below(s1, s2) {
    return union(scale(s1, 1, 1, 0.5),
                 translate(scale(s2, 1, 1, 0.5),
                           0, 0, 0.5));
}
function below_frac(f, s1, s2) {
    return union(scale(s1, 1, 1, f),
                 translate(scale(s2, 1, 1, 1 - f),
                           0, 0, f));
}

function menger(s, n) {
    if (n === 1) {
       return s;
    } else {
       const f = 1/3;
       const m = menger(s, n - 1);
       const outer_row = before_frac(f, m,
                               before(m, m));
       const inner_row = before_frac(f, m,
                               before(empty, m));
       const outer_layer = beside_frac(f, outer_row,
                                 beside(inner_row,
                                        outer_row));
       const inner_layer = beside_frac(f, inner_row,
                                 beside(empty,
                                        inner_row));
                                
       return below_frac(f, outer_layer,
                         below(inner_layer, 
                               outer_layer));
    }
}

const menger_sponge = menger(c, 4);

render(menger_sponge);

```

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


src/bundles/csg/functions.ts
```diff
@@ -181,6 +178,15 @@ export const white: string = '#FFFFFF';
...
 
+/**
+ * An empty Shape.
+ *
+ * @category Primitives
+ */
+export function empty_shape(): Shape {
+  return new Shape();
+}
+
```
martin-henz, line 189: Why is `empty_shape` a nullary function? Why can't it be just a shape value? I think it should be just called `empty`, and not `empty_shape`. Also, cube is just `cube` and not `cube_shape`.

martin-henz (changes requested):

src/bundles/csg/functions.ts
joeng03: Exported a function here to standardize with other primitives, which are all functions. I was feeling that `empty` probably doesn't provide that much context to the readers as to what this is?

martin-henz: I'm ok with calling it `empty_shape` but it should be a shape value, not a function. This would be similar to the runes `heart`, `square` etc in the `rune` module. BTW, there we call it `blank`. Should we call it `blank` here too?

Cloud7050: The primitive shapes used to all be values, but changes to the way users specified colours meant they all became functions, e.g. `cube(cyan)`. Therefore all currently provided shapes are functions:

image

As an empty shape wouldn't have a colour, it is true it does not need to be a function, but for consistency perhaps it would be neater to keep it as a nullary function. The only values CSG currently provides are colour strings, so if we were to export an `empty` value, it would be "on its own" as a shape (in its own category in the [documentation](https://source-academy.github.io/modules/documentation/modules/csg.html)).

As for naming, I'm personally leaning towards `empty` over `empty_shape`, and `blank` feels like it better fits runes over shapes.

martin-henz: The situation in `csg` is quite similar to `rune`. We have primitives (including an empty special case) and a way to color things. The two libraries go different routes in this. I feel that consistency between the two is more important than which way we go. So I'll approve this one, and make a separate issue regarding consistency between `rune` and `csg`.

martin-henz (approved):

PR #296, opened 2024-03-20
Scripts Refactor

leeyi45:
The first rewriting of the module scripts left a lot to be desired. With the benefit of new packages and lots of hindsight, here is a streamlined set of scripts that should resolve the issues where linting errors were not being properly caught by the CI.

Currently there's a really weird bug that continues to escape me with the testing for the json command, so I have that disabled. It's not really too important this instant, but it would be nice to figure out why Jest isn't calling the mocked function.

This refactor also includes some changes to how modules are loaded and how the JSON documentation is stored. I'll have the corresponding PRs in `js-slang` up.


devserver/src/components/Playground.tsx
```diff
@@ -16,7 +17,6 @@ import type { SideContentTab } from "./sideContent/types";
 import { getDynamicTabs } from "./sideContent/utils";
 
 // Importing this straight from js-slang doesn't work for whatever reason
```
RichDom2185, line 19: Can we move this comment too? (I know the reordering was automatic due to lint rules)

scripts/jest.config.js
```diff
@@ -0,0 +1,20 @@
...
+  testEnvironment: 'node',
+  moduleNameMapper: pathsToModuleNameMapper(tsconfig.compilerOptions.paths, { prefix: "<rootDir>/" }),
+  preset: 'ts-jest/presets/default-esm',
+  setupFilesAfterEnv: ["<rootDir>/jest.setup.ts"],
+  testMatch: [
+    '<rootDir>/src/**/__tests__/**/*.test.ts',
+  ],
+}
+
+export default jestConfig
```
RichDom2185, line 20: Can we add a newline at the end of this file?

scripts/jest.config.js
```diff
@@ -0,0 +1,20 @@
+import { pathsToModuleNameMapper } from 'ts-jest'
+import tsconfig from './tsconfig.json' assert { type: 'json' }
```
RichDom2185, line 2: Since we are using import assertions, we should probably bump the TypeScript version in `package.json` to `^5.3.0`.

scripts/jest.setup.ts
```diff
@@ -0,0 +1,30 @@
...
+    test1: { tabs: [] },
+    test2: {
+      tabs: ['tab1'],
+    },
+  })),
+}));
+
+global.process.exit = jest.fn(code => {
+	throw new Error(`process.exit called with ${code}`)
+})
```
RichDom2185, line 30: Ditto on newline at EOF.

RichDom2185 (commented):
Just did a quick run through – perhaps we can add "newline at EOF" as an error rule in `.eslintrc.base.cjs`?

leeyi45:
With the new flat configurations that ESLint is moving towards, I basically redid the ESLint setup for the repository. It should be far less janky than the way it had been setup previously and should play nicely with the VSCode extension (if you use the defaults provided in the .vscode folder)

src/bundles/scrabble/__tests__/index.ts
RichDom2185, line 1: Are we using 4 spaces now instead of 2 spaces for indentation?

RichDom2185: Why are tests using 4 spaces but other files using 2 spaces?

eslint.config.js
```diff
@@ -0,0 +1,732 @@
...
+  // "id-match": 0,
+  // "init-declarations": 0,
+  // "max-classes-per-file": 0,
+  // "max-depth": 0,
+  // "max-lines": 0,
+  // "max-lines-per-function": 0,
+  // "max-nested-callbacks": 0,
+  // "max-params": 0,
+  // "max-statements": 0,
+  // "multiline-comment-style": 0,
```
RichDom2185, line 166: These came from overrides from the old (now removed) AirBnB preset right? Can we just remove them? In fact I think might be best to purge all the overrides and start from a more modern preset like `@typescript-eslint/recommended` (or even better, `@typescript-eslint/recommended-type-checked`)

leeyi45: I'll go through everything and see if I can streamline the configuration

src/bundles/arcade_2d/gameobject.ts
```diff
@@ -1,6 +1,7 @@
 /**
  * This file contains the bundle's representation of GameObjects.
  */
+import Phaser from 'phaser';
```
RichDom2185, line 4: I'm trying to find where this import is used but I can't find it (and yet VSCode shows that it is not an unused import)

RichDom2185 (changes requested):
Semicolons are still not checked (some files have them, some files don't, and some have a mix)

src/bundles/arcade_2d/gameobject.ts
leeyi45: I am confused by this one to be honest

martin-henz:
The current sourceacademy.org frontend needs this PR to run modules.

RichDom2185 (approved):
LGTM, thanks!

PR #299, opened 2024-03-26
Linting fix

leeyi45:
The lint command wasn't properly exiting with error code 1 while linting errors remained. This PR fixes that and introduces tests to make sure of it.

Also put the `**/gl` resolution back into `package.json`


eslint.config.js
```diff
@@ -9,26 +9,18 @@ import tseslint from 'typescript-eslint';
 const todoTreeKeywordsWarning = ['TODO', 'TODOS', 'TODO WIP', 'FIXME', 'WIP'];
 const todoTreeKeywordsAll = [...todoTreeKeywordsWarning, 'NOTE', 'NOTES', 'LIST'];
 
-const OFF = 0;
```
martin-henz, line 12: This change appears to go in the opposite direction to what I would expect: You are replacing nicely factored out constants (numbers) by "magic strings": 'warn', 'off', 'error'. Why can't these remain constants?

leeyi45: I understand that if we use 0,1, or 2 they'd be magic constants, but ESLint lets directly use `off`, `warn` and `error`, which are self-explanatory. I don't see the need for this extra indirection.

martin-henz: Yes, I agree. Not much difference between having strings and identifiers that refer to strings.

RichDom2185 (approved):
LGTM, thanks!

PR #334, opened 2025-02-24
Upgrade to TS 5.8, ESLint v9, Vite v6 and other dependencies

RichDom2185:
# Description

Just the occasional maintenance. Also fixes dependency issues with NodeJS 22 on Windows


JustATin555 (changes requested):
Looks good, seems to introduce 3 "Unused eslint-disable directive" warnings which should be easily fixable with a `yarn lint --fix`.

JustATin555 (approved):

PR #335, opened 2025-02-26
Update unittest module to work with module contexts

leeyi45:
@RichDom2185 Apologies what I had meant was that the concept was basically ready for use, but the `testing` module does need a bit more work to be functional with the module context system.

I've also renamed the module to `unittest`; I think that's better than just `testing`


RichDom2185 (approved):
LGTM, thanks a lot!

PR #340, opened 2025-04-01
Robot minigame

JustATin555:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

- [ ] New and existing unit tests pass locally with my changes


src/tabs/RobotMaze/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  toSpawn(context: DebuggerContext) {
+    return context.context?.moduleContexts?.robot_minigame.state.isInit;
+  },
+
+  /**
+   * This function will be called to render the module tab in the side contents
+   * on Source Academy frontend.
+   * @param {DebuggerContext} context
+   */
+  body: (context: any) => <RobotMaze context={context} />,
```
RichDom2185, line 45: So is `context` a `DebuggerContext` or `any`?

src/tabs/RobotMaze/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * @author Koh Wai Kei
+ * @author Justin Cheng
+ */
+
+/**
+ * React Component props for the Tab.
+ */
+interface MainProps {
+  children?: never
+  className?: never
```
RichDom2185, line 16: Are these `never` needed?

src/tabs/RobotMaze/RobotSimulation.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+// The speed to move at
+const ANIMATION_SPEED : number = 2;
+
+/**
+ * React Component props for the Tab.
+ */
+interface MapProps {
+  children?: never
+  className?: never
```
RichDom2185, line 127: Ditto on `never`

src/bundles/robot_minigame/helpers/tests.ts
```diff
@@ -0,0 +1,34 @@
...
+    // Store status in a variable
+    let success: boolean;
+
+    switch(test.type) {
+      case 'area':
+        success = test.test(areaLog);
+        break;
+      default:
+        success = true;
+    }
```
RichDom2185, line 26: There's only one case, so a ternary suffices and is more readable (plus you get to define `success` as a const:

```suggestion
    // Store status in a variable
    const success = test.type === 'area' ? test.test(areaLog) : true;
```

Of course, representing it this way would make it more obvious that it's essentially a `||` given how short-circuit eval works:

```ts
    // Store status in a variable
    const success = test.type !== 'area' || test.test(areaLog);
```

Though this is less readable so I prefer the first suggestion

RichDom2185 (changes requested):
Did a quick run through as I'm outside right now, so just some preliminary comments, thanks!

src/bundles/robot_minigame/helpers/tests.ts
JustATin555: Wrote it this way for extensibility (as there may be other forms of checks), but I don't mind this suggestion either.

src/tabs/RobotMaze/index.tsx
JustATin555: Looks like a leftover from before my rewrite from class components to functional components.

JustATin555: Wait actually, seems to be present in other modules too, maybe a legacy thing? Will remove and see if it causes any errors.

lhw-1 (commented):
More of an overarching question - do you think it makes sense to frame it just as a maze minigame rather than a robot minigame? The idea of having it be a robot seems more context dependent within SA game, but as a module (which is published for not just Source Academy but for Source as a whole), it might be a nice idea to frame it as a general purpose maze minigame.

(Asking because this would necessitate some variable name changes)

src/tabs/RobotMaze/RobotSimulation.tsx
```diff
@@ -0,0 +1,287 @@
...
+  ctx : CanvasRenderingContext2D,
+  width : number,
+  height : number,
+  areas: Area[],
+  robot : Robot
+) => {
+  ctx.reset();
+  drawBorders(ctx, width, height);
+  drawAreas(ctx, areas);
+  drawRobot(ctx, robot);
```
lhw-1, line 116: Does it make sense to allow some optional parameters (e.g. colors, stroke style)?

src/tabs/RobotMaze/RobotSimulation.tsx
```diff
@@ -0,0 +1,287 @@
...
+};
+
+// The speed to move at
+const ANIMATION_SPEED : number = 2;
```
lhw-1, line 120: Might be nice to allow this to change as well

lhw-1 (commented):
Some small comments as well:

src/bundles/maze/src/functions.ts
```diff
@@ -0,0 +1,597 @@
+import { GeneralRuntimeError } from '@sourceacademy/modules-lib/errors';
+import context from 'js-slang/context';
+// import {
```
AaravMalani: Would recommend removing the comment

src/bundles/maze/src/functions.ts
```diff
@@ -0,0 +1,597 @@
...
+  flags: [key: string, value: any][]
+) {
+  // Init functions should not run after initialization
+  if (state.isInit) throw new GeneralRuntimeError('May not use initialization functions after initialization is complete!');
+
+  // Store vertices as Point array
+  const parsedVertices: Point[] = vertices.map(v => ({ x: v[0], y: v[1] }));
+
+  // Store flags as an object
+  const parsedFlags = flags.reduce((acc, f) => ({ ...acc, [f[0]]: f[1] }), {});
```
AaravMalani: Would it make sense to use `Object.fromEntries` here? Might make the code a bit more readable

src/bundles/maze/src/functions.ts
```diff
@@ -0,0 +1,597 @@
...
+ * @returns if the robot entered the given colors in order
+ */
+export function should_enter_colors(
+  colors: string[]
+) {
+  state.tests.push({
+    type: 'area',
+    test: (areas: Area[]) => {
+      const coloredAreas = areas
+        .filter((area: Area) => colors.includes(area.flags.color)) // Filter relevant colors
```
AaravMalani: Is the type annotation necessary here?

AaravMalani (commented):
Minor stuff, should be fine

AaravMalani:
@coderabbitai review

PR #465, opened 2025-06-23
Clean up some tabs

RichDom2185:
# Description

To merge to #459


leeyi45:
Should we enforce using functional components only? LGTM otherwise

PR #481, opened 2025-11-05
Yarn Workspaces Follow Up

### LINKED ISSUE #478: Root package missing in topo order?
Image

Seems like that's the cause of all the dependency updates failing (since they only touch the lockfile)

leeyi45:
This PR is intended to fix the little teething issues here and there, especially with the CI:

- Incorrect docserver base path resulting in the styling and links being all wrong
- Sneaky bug in the github actions workflow that was throwing errors on certain packages (fixes #478)
- Devserver not being able to build the tabs that it needs to build
- The matrix jobs still trying to spin up a whole job just to stop execution when their corresponding package doesn't have changes
- The matrix jobs trying to execute even if there aren't any tabs, bundles or libraries that have changes
- Certain jobs not executing if the matrix jobs don't execute at all


leeyi45:
At first glance, it does seem that the CI actions do work. At the time of writing when I only have minor changes in the docs server it isn't trying to run every other job

PR #484, opened 2025-11-10
Detect Changes based on Lockfile

leeyi45:
@RichDom2185 provided some code that would help detect which workspaces need to be rebuilt based on diffs across lockfiles. This PR will implement that change, as well as fix some typos in the documentation.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11976,14 +11976,14 @@ __metadata:
   linkType: hard
 
 "js-yaml@npm:^3.10.0, js-yaml@npm:^3.13.1":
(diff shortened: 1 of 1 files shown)


leeyi45:
Also no need to merge just yet, there are probably other things I need to add

RichDom2185:
There seems to be duplicate computation with the current workflow because when you build tabs, you need to build bundles. Could we just remove them?

leeyi45:
Superseded by #519, but there are changes here that should probably be revisited

PR #519, opened 2025-12-04
Transitive Context Import Fix

### LINKED ISSUE #518: [PLOTLY]: examples throws "Errors occurred when executing the module"
# Expected Behavior

Plot should appear in Plotly tab

# Current Behavior

Plot does not appear in Plotly tab, and shows `Error(s) occured when executing the module 'plotly'.` in the REPL.

# Failure Information (for bugs)

## Steps to Reproduce

```javascript
import { draw_connected_2d } from "plotly";
```

1. Go to [this example](https://share.sourceacademy.org/uolem) from [docs](https://source-academy.github.io/modules/documentation/functions/plotly.new_plot.html).
    - or navigate to [this example](https://share.sourceacademy.org/tx4ba) from [this page](https://source-academy.github.io/modules/documentation/functions/plotly.draw_connected_2d.html)
    - or just import from plotly like above
2. Run with Source 4 or full javascript
3. See error

## Context

Error does not appear on [SA @ NUS](https://share.sourceacademy.nus.edu.sg/m8g84).

Tested on

- Version used: `a58d1618bfda6cf57b7f2971a822f10a9ab0f642`
- Browser Name and version: Firefox for NixOS 145.0.1 (64-bit), Chrome Version 142.0.7444.175 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): NixOS 25.11.20251124.a320ce8 (Xantusia)

and

- Version used: `a58d1618bfda6cf57b7f2971a822f10a9ab0f642`
- Browser Name and version: Chrome Version 142.0.7444.175 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): Linux Mint 22.1

## Failure Logs

No error shown in console.

leeyi45:
#518 highlighted that if bundles relied on one another as dependencies, as `plotly` does with `sound` and `curve`, importing the usual way would cause the context import to get triggered.

Because `js-slang` doesn't load the contexts for bundles that are loaded transitively, the module context for the imported bundle is `undefined` and setting its value raises a `TypeError`. Specifically in the case of `plotly`:

```ts
// plotly/functions.ts

// this line triggers the evaluation of curve/functions.ts
import { x_of } from '@sourceacademy/bundle-curve';
```

```ts
// curve/functions.ts
import context from 'js-slang/context';
const drawnCurves = [];

// This line throws a TypeError when evaluated because the curve
// bundle isn't loaded directly by js-slang
context.moduleContexts.curve.state = {
  drawnCurves
};
```

The fix is simple: move functions that are context dependent to their own file and make sure not to import from that file.

This PR adds an ESBuild plugin to detect when this is happening and also refactor some of the bundles and imports to prevent this from happening, but prohibiting this kind of import doesn't necessarily have to be the solution since it will also prohibit relying on context dependent functions from another bundle (what if `plotly` did want to use `draw_connected` from `curve`?)

(Should close #518)

## Other Miscellanous Bug Fixes/Features
- Fixed a potential bug where a Windows root path might not be recognized as and converted to a root path
- Improved documentation around mocking
- Added highlighting to dirtree diagrams
- Use @RichDom2185's implementation for the lockfile checks

### DIFF
--- a/docs/src/modules/1-getting-started/4-cheat.md
+++ b/docs/src/modules/1-getting-started/4-cheat.md
@@ -34,7 +34,7 @@ All commands have a `-h` or `--help` option that can be used to get more informa
     <tr>
       <td><code>yarn add -D &lt;package&gt;</code></td>
       <td>
-        Adds a package to the current bundle or tab that will is not required at runtime <br />
+        Adds a package to the current bundle or tab that is not required at runtime <br />
       </td>
     </tr>
   </tbody>


RichDom2185 (approved):
LGTM, thank you!

PR #607, opened 2026-03-03
Better Type Errors

leeyi45:
# Better Error Handling and Documentation

Some general fixes to things that got missed out in previous PRs. Also re-exports `js-slang`'s rttc (runtime type checking) utilities and exposes them to modules

## Allowing Rest, Default and Optional Parameters

If wrapped appropriately with `wrap`, module functions can now have variable numbers of arguments.

## `throw-runtime-error`
Introduces a linting rule as well to highlight whenever an error being thrown doesn't inherit from the RuntimeSourceError type.

## Devserver Changes

Since `js-slang` now allows us to customize what importers are being used, the devserver can now be used without the modules server.

## `copy_gc` and `mark_sweep` refactor

I modernized those two modules to bring them up to date with module contexts and other design patterns. They are sorely lacking in documentation however, so I don't actually know if anything is broken.

## `noImplicitAny`

`noImplicitAny` has been enabled for all bundles and tabs and only left disabled for modules (like `unity_academy`) that simply don't leave enough context clues or documentation for me to be able to create types for them. I've fixed type errors wherever I can, but a lot of bundles don't have tests so it's not always clear if any kind of functionality has been affected.

## Documentation Changes

Fixed the following bugs:
1. The buildtools weren't properly detecting variables that were actually functions
2. Default values for constants were being rendered with their initializing expression, thus leaking implementation details.
3. buildtools sometimes didn't render type guard returns types correctly

To fix all of these problems I just gave in and migrated most of the Typedoc logic into its own plugin and package.


leeyi45:
Something very weird is happening with this PR

Across the three different environments that I have access to (Windows, MacOS and the Github Actions) everything seems to be behaving inconsistently

1. Tests that are succeeding on Windows are failing in MacOS and Github Actions but for different reasons
2. The import/extensions error seems to only be reported in Github Actions
3. A ton of the Typescript errors are just not being reported on Windows

This'll probably warrant more investigation on my end....

leeyi45:
I am still considering introducing the validator in this PR, but I think I will probably have to leave it for another PR. There's a bunch of design decisions for it that need to be made.

leeyi45:
Certain ESLint rules that rely on Typescript linting also don't work on the CI but work locally. There's some weird shennenigans going on with the Typescript present in the CI environment and the in the local environment.

RichDom2185 (approved):
LGTM, thank you!

PR #640, opened 2026-04-06
Migrate to Typescript 6

leeyi45:
Migrate everything in the repository to Typescript 6 and update some dependencies.

For some reason the ESLint 10 config algorithm doesn't work, so I've left off on updating ESLint to V10.

Also fix some code that wasn't always behaving properly.


leeyi45:
There is a type error with the wasm bundle, not sure what to do about that one.

leeyi45:
@RichDom2185 I've figured out what the Yarn error was.

The error was occurring with the running of the `yarn why` command. When you run `yarn why` for Typescript 6, one of the entries that comes up has a locator that looks like this: `typescript@patch:typescript@npm:6.0.2`. The `extractPackageName` function then returns `typescript@patch:typescript`, which of course `yarn why` can't process and thus throws an error for.

The error is hidden because the code that runs `yarn why` looks like this:

```ts
  const { stdout: output, exitCode, stderr } = await getExecOutput('yarn', ['why', pkgName, '--json'], { silent: true });
  if (exitCode !== 0) {
    core.error(stderr);
    throw new Error(`yarn why for ${pkgName} failed!`);
  }
```

The `silent: true` hides any `stdout` output (including the command itself). Because `failOnStdErr` and `ignoreReturnCode` are also both unset, the fact that the command exits with return code 1 causes `getExecOutput` to throw an error, bypassing the exit code check entirely.

The combined outcome is command is essentially failing silently.

lib/lintplugin/src/configs.ts
```diff
@@ -180,6 +180,7 @@ export const tsxConfig = defineConfig({
   name: 'SA TSX Config',
   files: ['**/*.tsx'],
   plugins: {
+    // @ts-expect-error React hooks plugin wrong type
     'react-hooks': reactHooksPlugin,
```
RichDom2185, line 184: If we're updating to v7, then we should use `reactHooks.configs.flat['recommended']` and then just disable unneeded rules in a subsequent block, no manual plugin registration needed.

RichDom2185 (changes requested):
A number of ESLint plugins (notably React) don't support v10 yet so I'm not surprised things are failing, but I didn't really look into it since should we even look into updating ESLint to v10 when the rest of the ecosystem hasn't caught up?

lib/lintplugin/src/configs.ts
leeyi45: The plugin is no longer a plugin?

RichDom2185: It does seem like the default export still exports the plugin itself (based on their docs); I guess this is fine though with flat config I feel like we should break it up into hierarchies instead of large objects, hence the suggestion.

RichDom2185:
> There is a type error with the wasm bundle, not sure what to do about that one.

Sorry, just had time to look at it. Just need to inline it, I'm also not sure what's going on with the type inference but that's how I've been fixing it when I encountered the buffer-related issues.

.github/actions/src/lockfiles.ts
```diff
@@ -169,5 +173,6 @@ export const hasLockFileChanged = memoize(async () => {
       ignoreReturnCode: true
     }
   );
+  core.info('i finished executing');
```
RichDom2185, line 176: Haha let's keep this as an Easter Egg

lib/lintplugin/src/configs.ts
leeyi45: I prefer that each plugin is listed explicitly? It helps highlight exactly which configs have which plugins loaded.

RichDom2185: Fair, let's keep it this way then

.github/actions/src/lockfiles.ts
leeyi45: ☠️

RichDom2185 (approved):
LGTM, thanks a lot for working on this!

PR #680, opened 2026-04-16
Deploy conductor-ready modules to modules-conductor 

martin-henz:
Deploys build artifacts from the conductor-migration branch to source-academy/modules-conductor via GitHub Pages, allowing the conductor version of modules to coexist with the current deployment.


RichDom2185 (changes requested):
I don't get the purpose of this PR, is it for CD? What's left to get it out of draft status?

martin-henz:
> I don't get the purpose of this PR, is it for CD? What's left to get it out of draft status?

Every push to this branch updates the modules-conductor repo. So yea, it's for CD in that sense.

Having a PR allows us to easily merge any commits to the main branch into this branch, to keep modules-conductor up-to-date.

RichDom2185:
> Having a PR allows us to easily merge any commits to the main branch into this branch, to keep modules-conductor up-to-date.

I agree, but this sounds like this is a PR for the development trunk instead of CD then. In which case it would remain in draft forever and the workflow never get merged or run?

PR #683, opened 2026-04-24
Update README.md to reflect conductor migration

martin-henz:
adding section on conductor migration

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,4 +51,4 @@ You should now have a local copy of the modules' developer documentation running
 
 ## Conductor migration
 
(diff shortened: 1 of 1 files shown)


RichDom2185 (changes requested):
The workflow for the PR won't run if it's not merged. So there should be no reference to the PR status and just reference the file directly. This should not be a separate PR

RichDom2185:
> The workflow for the PR won't run if it's not merged. So there should be no reference to the PR status and just reference the file directly. This should not be a separate PR

If the other PR is meant to be a development trunk, then the workflow file should be here so it can get merged to main. Otherwise it will never run

PR #765, opened 2026-07-06
Migrate Rune Module

AaravMalani:
# Description

This PR migrates the `rune` module to become Conductor-ready. This includes adding serialisation and deserialisation methods and a plugin/channel transport mechanism. It also updates the buildtools documentation engine to support the `functionDeclaration` and `variableDeclaration` decorators for type inference.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [x] Running the module on the Conductor-enabled Source Academy
- [x] Running `yarn test` with updated tests

# Checklist:


src/bundles/rune/src/rune.ts
```diff
@@ -223,7 +223,22 @@ export function drawRunesToFrameBuffer(
...
+          src: imageSource
+        });
+        return new Promise((resolve, reject) => {
+          image.onload = () => resolve(image);
+          image.onabort = reject;
+          image.onerror = reject;
+        });
+      }
+      return imageSource;
+    })();
```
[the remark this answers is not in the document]
AaravMalani: Fixed

src/tabs/Rune/src/index.tsx
```diff
@@ -1,45 +1,180 @@
...
+    const rune = deserializeRune(this.message.frames[frame]);
+    const drawnRune = this.message.mode === 'anaglyph'
+      ? new AnaglyphRune(rune)
+      : new NormalRune(rune);
+
+    return {
+      draw: drawnRune.draw
+    };
+  }
+}
```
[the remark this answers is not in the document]
AaravMalani: I don't agree with this, especially if there are 1000+ frames. I think the memory tradeoff isn't worth it
I'm leaving this up to the reviewer to decide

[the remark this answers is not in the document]
leeyi45: hmm. In practice, I've never really noticed significant performance degradation having the runes being drawn like this. I think this is okay, but we should probably move all rendering code into the tab eventually. From there if there are optimizations to be made then we can do that.

I also avoided memoizing the rendered runes so that you could have non-deterministic animations. We could decide we don't want that, but I think adding the memoization adds a layer of complexity that challenges how a cadet might expect a function to behave (i.e calling `math_random` multiple times ought to result in random behaviour)

In other words, if there's no significant impact to performance, I'm also okay to leave it this way,

lib/buildtools/src/build/docs/conductor/declarations.ts
leeyi45, line 1: I do implore you to add some explanation on what each function is trying to achieve here, and thus what kind of output is expected, or what each `interface` is intended to be

lib/buildtools/src/build/docs/conductor/declarations.ts
```diff
@@ -0,0 +1,209 @@
...
+      metadata.variableDeclaration = getStringArgument(call, 0);
+    }
+  });
+
+  return metadata;
+}
+
+function parseFunctionDeclaration(paramTypes: string, returnType: string, sourceName: string): FunctionDeclarationMetadata {
+  const declarationSource = `function __sourceAcademyDocs__(${paramTypes}): ${returnType} {}`;
+  const sourceFile = ts.createSourceFile(`${sourceName}.functionDeclaration.ts`, declarationSource, ts.ScriptTarget.Latest, true);
```
leeyi45: Do we have to get Typescript to do the parsing? I mean it seems like we know all the details, so we could just populate the metadata object ourselves

src/bundles/rune/src/index.ts
```diff
@@ -1,63 +1,997 @@
...
+import { Rune } from './rune';
+import { throwIfNotRune } from './runes_ops';
+import { functionDeclaration, type_map, variableDeclaration } from './type_map';
+
+type RuneTabLoader = {
+  tabs: string[];
+  loadTab: (tab: string) => void;
+};
+
+const RUNE_CONSTANTS = [
```
leeyi45: I think this value should be defined by the actual rune constants and not just be defined on its own. You can probably get these values using `Object.getOwnPropertyNames`.

And then you can do some type shennenigans like:

```ts
type RuneConstantNames = {
  [K in keyof typeof RuneConstants as (typeof RuneConstants)[K] extends () => Rune ? K : never]: K
}

type RuneConstantFunctions = keyof RuneConstantFunctions;
```

src/bundles/rune/src/index.ts
```diff
@@ -1,63 +1,997 @@
...
+    const frameCount = Math.max(1, Math.ceil(duration * fps) + 1);
+    const frames: RuneAnimationMessage['frames'] = [];
+
+    for (let frame = 0; frame < frameCount; frame += 1) {
+      const timestamp = Math.min(frame / fps, duration);
+      const result = yield* this.evaluator.closure_call_unchecked(
+        func as TypedValue<DataType.CLOSURE, DataType.OPAQUE>,
+        [{ type: DataType.NUMBER, value: timestamp }]
+      );
+      frames.push(serializeRune(await this.__getRune(result, funcName)));
```
leeyi45, line 283: This does kind of have me wondering, would it make sense to just render the entire animation at once with `Promise.all`?

src/bundles/rune/src/index.ts
```diff
@@ -1,63 +1,997 @@
...
+  returnType: DataType
+) {
+  const method = RuneModulePlugin.prototype[methodName] as SignaturedModuleMethod | undefined;
+  if (method === undefined) {
+    throw new Error(`Rune module method "${methodName}" does not exist.`);
+  }
+  method.signature = { args, returnType };
+}
+
+attachModuleMethod('anaglyph', [DataType.OPAQUE], DataType.OPAQUE);
```
leeyi45: I am always wary of cases where there are multiple sources of truth. In this case, there's some type games we can play to force the attachment call to match the definitions of the actual function:

```ts
type TypeToDataType<T> =
  T extends any[]
    ? T['length'] extends 2
      ? DataType.PAIR | DataType.ARRAY
      : DataType.ARRAY
    : T extends (...args: any[]) => any
      ? DataType.CLOSURE
      : T extends boolean
        ? DataType.BOOLEAN
        : T extends null
          ? DataType.EMPTY_LIST
          : T extends number
            ? DataType.NUMBER
            : T extends string
              ? DataType.CONST_STRING
              : DataType.OPAQUE;

type TupleToDataTypes<T extends any[]> =
  T extends []
    ? []
    : T extends [infer U, ...infer Rest]
      ? [TypeToDataType<U>, ...TupleToDataTypes<Rest>]
      : never;

function attachModuleMethodNew<T extends (...args: any[]) => any>(
  f: T,
  args: TupleToDataTypes<Parameters<T>>,
  returnType: TypeToDataType<ReturnType<T>>
) {
  const methodName = f.name;
  const method = RuneModulePlugin.prototype[methodName] as SignaturedModuleMethod | undefined;
  if (method === undefined) {
    throw new Error(`Rune module method "${methodName}" does not exist.`);
  }
  method.signature = { args, returnType };
}

attachModuleMethodNew(funcs.beside, [DataType.OPAQUE, DataType.OPAQUE], DataType.OPAQUE);
attachModuleMethodNew(funcs.color, [DataType.OPAQUE, DataType.NUMBER, DataType.NUMBER, DataType.NUMBER], DataType.OPAQUE);
```

The helper types here probably need more nuance, and should honestly live outside of the rune bundle altogether, but you get the gist.

leeyi45: I guess the wider question would be if all of this can be avoided, but as an interim measure during this transition period it's okay.

src/bundles/rune/src/rune.ts
```diff
@@ -223,7 +223,27 @@ export function drawRunesToFrameBuffer(
    * Initialize a texture and load an image.
    * When the image finished loading copy it into the texture.
    */
-  const loadTexture = (image: HTMLImageElement): WebGLTexture | null => {
+  const loadTexture = async (rune: Rune): Promise<WebGLTexture | null> => {
```
leeyi45, line 229: I would rather have it written like this:

```ts
    let image: HTMLImageElement;
    if (typeof imageSource !== 'string') {
      image = imageSource;
    } else {
      image = await new Promise<HTMLImageElement>((resolve, reject) => {
        const image = Object.assign(new Image(), {
          crossOrigin: 'anonymous',
          src: imageSource
        });
        image.onload = () => {
          rune.texture = image;
          resolve(image);
        };
        image.onabort = reject;
        image.onerror = reject;
      });
    }
```
than have an IIFE

src/tabs/Rune/src/hollusion_canvas.tsx
```diff
@@ -13,14 +13,21 @@ type Props = {
...
+  const renderFuncRef = React.useRef<(time: number) => unknown>(undefined);
+  const isLoading = React.useRef(false);
   const { setCanvas } = useAnimation({
     callback(timestamp, canvas) {
-      if (renderFuncRef.current === undefined) {
-        renderFuncRef.current = rune.draw(canvas);
+      if (renderFuncRef.current) {
+        return renderFuncRef.current(timestamp);
+      }
+      if (!isLoading.current) {
```
leeyi45: Is the `isLoading` ref necessary? It seems like we could just share the `renderFuncRef`. Something like

```ts
type RenderFuncState = 'loading' | ((time: number) => unknown)> | null;
const renderFuncRef = React.useRef<RenderFuncState>(null);
```

src/tabs/Rune/src/index.tsx
```diff
@@ -1,45 +1,180 @@
...
+
+function RenderedRune({ message }: { message: Extract<RuneDisplayMessage, { type: 'render' }> }) {
+  const rune = useMemo(() => deserializeRune(message.rune), [message]);
+  const drawnRune = useMemo(() => {
+    if (message.mode === 'anaglyph') return new AnaglyphRune(rune);
+    if (message.mode === 'hollusion') return new HollusionRune(rune, message.magnitude ?? 0.1);
+    return new NormalRune(rune);
+  }, [message, rune]);
+
+  if (drawnRune instanceof HollusionRune) {
```
leeyi45: Have you verified that this `instanceof` check works? Because of the way tabs and bundles are compiled this check might fail at runtime. See [here](https://source-academy.github.io/modules/devdocs/modules/5-advanced/misc.html#instanceof-will-not-work-at-runtime)

AaravMalani: It does work. Earlier, the tabs and bundles shared the same object. Now since it's deserialised on the tab side with the tab's HollusionRune class, the `instanceof` check works

src/bundles/rune/src/rune.ts
AaravMalani: Works! Let me change it

src/bundles/rune/src/index.ts
AaravMalani: Yep, I agree that replicating it twice is a pain, but I'm not convinced the tower of types is the best way.
Let me get back to you and see what is possible

lib/buildtools/src/build/docs/conductor/declarations.ts
AaravMalani: Okay, so I'm looking at Typedoc, and according to a GH issue from a year ago
> The support which TypeDoc removed was for parsing real decorators and removing them from the output, not the tag. It predated support for real decorators and was never actually used by any of TypeDoc's code.

So I think I'm gonna migrate to using an `@public` tag to provide the declaration info
Does that sound good?

src/bundles/rune/src/index.ts
AaravMalani: Fair enough, I'll make the change

AaravMalani: > I think this value should be defined by the actual rune constants and not just be defined on its own. You can probably get these values using `Object.getOwnPropertyNames`.
> 
> And then you can do some type shennenigans like:
> 
> ```ts
> type RuneConstantNames = {
>   [K in keyof typeof RuneConstants as (typeof RuneConstants)[K] extends () => Rune ? K : never]: K
> }
> 
> type RuneConstantFunctions = keyof RuneConstantFunctions;
> ```

Are the type shenanigans even required? I could just iterate over the `RuneFunctions` and add the runes to the class. Even if they don't line up with the class declaration, its okay because the class variables are only used for documentation

lib/buildtools/src/build/docs/conductor/declarations.ts
leeyi45: If we want to use a tag, I think we should create our own custom tag to do it.

But what I meant was that it looks as if you're creating a sample program, then getting Typescript to parse the AST data when you could probably skip that step and just go from the `paramTypes`, `returnType` and `sourceName` straight to the output.

AaravMalani: Yep, I've created a custom `@publicType` and `@publicReturnType` tag. Could you review the new pipeline?

src/bundles/rune/src/index.ts
AaravMalani: I've moved the `attachModuleMethod` function to module-lib!

src/tabs/Rune/src/hollusion_canvas.tsx
AaravMalani: Agreed, but I'd rather use { type: ...., value?: ... } the state instead

src/bundles/rune/src/index.ts
AaravMalani: I think that would cause a race condition, considering the CSE evaluator can only execute one function at a time. 

A later frame could push to the CSE control during CSE execution, which would mess the execution up.

src/bundles/rune/src/index.ts
```diff
@@ -1,63 +1,1050 @@
...
+attachModuleMethod<RuneModulePlugin, 'beside'>(RuneModulePlugin, 'beside', [DataType.OPAQUE, DataType.OPAQUE], DataType.OPAQUE);
+attachModuleMethod<RuneModulePlugin, 'beside_frac'>(RuneModulePlugin, 'beside_frac', [DataType.NUMBER, DataType.OPAQUE, DataType.OPAQUE], DataType.OPAQUE);
+attachModuleMethod<RuneModulePlugin, 'black'>(RuneModulePlugin, 'black', [DataType.OPAQUE], DataType.OPAQUE);
+attachModuleMethod<RuneModulePlugin, 'blue'>(RuneModulePlugin, 'blue', [DataType.OPAQUE], DataType.OPAQUE);
+attachModuleMethod<RuneModulePlugin, 'brown'>(RuneModulePlugin, 'brown', [DataType.OPAQUE], DataType.OPAQUE);
+attachModuleMethod<RuneModulePlugin, 'color'>(RuneModulePlugin, 'color', [DataType.OPAQUE, DataType.NUMBER, DataType.NUMBER, DataType.NUMBER], DataType.OPAQUE);
+attachModuleMethod<RuneModulePlugin, 'flip_horiz'>(RuneModulePlugin, 'flip_horiz', [DataType.OPAQUE], DataType.OPAQUE);
+attachModuleMethod<RuneModulePlugin, 'flip_vert'>(RuneModulePlugin, 'flip_vert', [DataType.OPAQUE], DataType.OPAQUE);
+attachModuleMethod(RuneModulePlugin, 'from_url', [DataType.CONST_STRING], DataType.OPAQUE);
```
leeyi45, line 1045: Do we really need the type arguments?

lib/modules-lib/src/conductor/methods.ts
```diff
@@ -0,0 +1,43 @@
+
+import type { IModulePlugin } from '@sourceacademy/conductor/module';
+import type { DataType, TypedValue } from '@sourceacademy/conductor/types';
+
+type ModuleFunction<
```
leeyi45: Add documentation for these types; they'll appear in VSCode Intellisense.

Also would it be possible to add type tests? Vitest supports `expectTypeOf`.

src/bundles/rune/src/index.ts
AaravMalani: Ah sorry, I forgot to remove them from testing

lib/modules-lib/src/conductor/methods.ts
AaravMalani: Added!

PR #790, opened 2026-07-11
Migrate binary_tree module to Conductor

Akshay-2007-1:
## Description

Fixes #768

Migrates the `binary_tree` module to be Conductor-ready, following the pattern established by the `repeat` module (#698). `binary_tree` has no tab/visualization component, so this is scoped entirely to `src/bundles/binary_tree/`.

- `functions.ts` is rewritten against Conductor's `IDataHandler` (`pair_make`/`pair_head`/`pair_tail`) instead of js-slang's `stdlib/list`, so tree structure is built from real Conductor Pairs rather than native JS arrays.
- `index.ts` is now a `BaseModulePlugin` subclass with each exported function as a `@moduleMethod`-decorated closure, mirroring `repeat`.
- The stored tree entry value is declared `DataType.OPAQUE` at the module boundary (arbitrary Source/Python values aren't representable as a single concrete `DataType`).
- `is_tree`/`is_empty_tree` declare no argument type, since they're predicates that must accept a value of *any* type and answer `false` rather than throw — a declared type would mean the FFI boundary type-asserts and throws before the predicate body ever runs.
- `tsconfig.json` picked up the same `experimentalDecorators`/`emitDecoratorMetadata` overrides `repeat` already has (this bundle was missing them, which breaks the `moduleMethod` decorator under TS's native Stage-3 decorators).

## Testing

- `yarn workspace @sourceacademy/bundle-binary_tree run tsc` — passes
- `yarn workspace @sourceacademy/bundle-binary_tree run lint` — passes
- `yarn workspace @sourceacademy/bundle-binary_tree run test` — 16/16 passing, against `@sourceacademy/modules-testplugin`'s `TestDataHandler`
- `yarn workspace @sourceacademy/bundle-binary_tree run build` — produces `build/bundles/binary_tree.js`
- End-to-end: loaded the built bundle into py-slang's `PyCseEvaluator` (via py-slang#217's module loader work) and drove `make_tree`/`entry`/`left_branch`/`right_branch`/`is_tree` from real Python-shaped values through the actual `pythonToModule`/`moduleToPython` conversion code — all round-trip correctly.

## Notes for reviewers

Getting the end-to-end test working surfaced two bugs upstream of this PR, not fixable within `binary_tree` itself:
- `BaseModulePlugin.initialise()` in `conductor` registers each exported method as a detached function reference (`this[name]`) without binding it to the instance, so `this` is `undefined` inside any exported method body once a real evaluator calls it. This breaks every Conductor module that uses ordinary class methods (i.e. all of them), not just this one. Filing a fix against `conductor` separately.
- `moduleToPython`/`pythonToModule` in py-slang#217 flatten a returned `DataType.PAIR` into a Python list and convert Python lists back to `DataType.ARRAY`, which loses identity for any function returning a Pair that's meant to be passed back into another module call (e.g. `left_branch(t)`). Flagging this on that PR directly since the affected file is new there.

### DIFF
--- a/src/bundles/binary_tree/src/functions.ts
+++ b/src/bundles/binary_tree/src/functions.ts
@@ -34,6 +34,14 @@ export async function make_tree(
   left: BinaryTree,
   right: BinaryTree
 ): Promise<NonEmptyBinaryTree> {
+  if (!await is_tree(evaluator, left)) {
+    throw new EvaluatorTypeError(`${make_tree.name} expects binary tree for left`, 'binary tree', DataType[left.type]);
+  }
+
+  if (!await is_tree(evaluator, right)) {
+    throw new EvaluatorTypeError(`${make_tree.name} expects binary tree for right`, 'binary tree', DataType[right.type]);
+  }
+
   const rightPair = await evaluator.pair_make(right, mEmptyList());
   const leftPair = await evaluator.pair_make(left, rightPair);
   return evaluator.pair_make(value, leftPair);
--- a/src/bundles/binary_tree/src/functions.ts
+++ b/src/bundles/binary_tree/src/functions.ts
@@ -51,6 +51,7 @@ export async function make_tree(
  * @param value Value to be tested. May be of any Conductor DataType.
  */
 export async function is_tree(evaluator: IDataHandler, value: TypedValue<DataType>): Promise<boolean> {
+  if (!value) return false;
   if (value.type === DataType.EMPTY_LIST) return true;
   if (value.type !== DataType.PAIR) return false;
 
@@ -82,7 +83,7 @@ export async function is_tree(evaluator: IDataHandler, value: TypedValue<DataTyp
  * @returns bool
  */
 export function is_empty_tree(value: TypedValue<DataType>): value is TypedValue<DataType.EMPTY_LIST> {
-  return value.type === DataType.EMPTY_LIST;
+  return value?.type === DataType.EMPTY_LIST;
 }
 
 async function assertNonEmptyTree(
@@ -90,8 +91,8 @@ async function assertNonEmptyTree(
   value: TypedValue<DataType>,
   funcName: string
 ): Promise<NonEmptyBinaryTree> {
-  if (!await is_tree(evaluator, value)) {
-    throw new EvaluatorTypeError(`${funcName} expects binary tree`, 'binary tree', DataType[value.type]);
+  if (!value || !await is_tree(evaluator, value)) {
+    throw new EvaluatorTypeError(`${funcName} expects binary tree`, 'binary tree', value ? DataType[value.type] : 'undefined');
   }
 
   if (value.type !== DataType.PAIR) {
--- a/src/bundles/binary_tree/src/functions.ts
+++ b/src/bundles/binary_tree/src/functions.ts
@@ -1,4 +1,6 @@
-import { head, is_list, is_pair, list, tail } from 'js-slang/dist/stdlib/list';
+import { EvaluatorRuntimeError, EvaluatorTypeError } from '@sourceacademy/conductor/common';
+import { DataType, type IDataHandler, type TypedValue } from '@sourceacademy/conductor/types';
+import { mEmptyList } from '@sourceacademy/conductor/util';
 import type { BinaryTree, EmptyBinaryTree, NonEmptyBinaryTree } from './types';
 
 /**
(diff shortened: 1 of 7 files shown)


src/bundles/binary_tree/src/index.ts
```diff
@@ -8,12 +8,78 @@
...
+    if (!value) throw new Error('is_tree expects 1 argument, received 0');
+    return { type: DataType.BOOLEAN, value: await is_tree_func(this.evaluator, value) };
+  }
+
+  // Same reasoning as is_tree: must accept a value of any Conductor DataType.
+  @moduleMethod([], DataType.BOOLEAN)
+  async* is_empty_tree(value?: TypedValue<DataType>): AsyncGenerator<void, TypedValue<DataType.BOOLEAN>, unknown> {
+    if (!value) throw new Error('is_empty_tree expects 1 argument, received 0');
+    return { type: DataType.BOOLEAN, value: is_empty_tree_func(value) };
+  }
```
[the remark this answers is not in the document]
Akshay-2007-1: @gemini-code-assist,
Does this mean the conductor framework has multiple error types like js-slang?

src/bundles/binary_tree/src/functions.ts
```diff
@@ -62,21 +87,27 @@ export function is_tree(value: any): value is BinaryTree {
...
  * display(is_empty_tree(tree)); // Shows "false" in the REPL
  * ```
- * @param value Value to be tested
+ * @param value Value to be tested. May be of any Conductor DataType.
  * @returns bool
  */
-export function is_empty_tree(value: BinaryTree): value is EmptyBinaryTree {
-  return value === null;
+export function is_empty_tree(value: TypedValue<DataType>): value is TypedValue<DataType.EMPTY_LIST> {
+  return value?.type === DataType.EMPTY_LIST;
```
AaravMalani, line 94: Out of curiosity, when would value be nullish?

src/bundles/binary_tree/src/index.ts
```diff
@@ -8,12 +8,100 @@
...
+    super(conduit, channels, evaluator);
+    this.__bindExportedMethods();
+  }
+
+  // BaseModulePlugin.initialise() registers each exported method as a detached `this[name]`
+  // reference without binding it, so any evaluator that calls the resulting closure as a bare
+  // function (which is how every evaluator calls closures) gets `this === undefined` inside the
+  // method body. Same workaround as repeat/rune until the upstream fix in conductor lands
+  // (source-academy/conductor#41).
+  private __bindExportedMethods() {
```
AaravMalani: Yep, can be removed now

src/bundles/binary_tree/src/index.ts
```diff
@@ -8,12 +8,100 @@
...
+  @moduleMethod([DataType.OPAQUE, DataType.LIST, DataType.LIST], DataType.PAIR)
+  async* make_tree(
+    value: TypedValue<DataType.OPAQUE>,
+    left: TypedValue<DataType.LIST>,
+    right: TypedValue<DataType.LIST>
+  ): AsyncGenerator<void, TypedValue<DataType.PAIR>, unknown> {
+    return await make_tree_func(this.evaluator, value, left, right);
+  }
+
+  // No declared arg type: is_tree is a predicate that must accept a value of any
```
AaravMalani, line 52: Hmm, for this I think we should add a `DataType.ANY` which just accepts all values. Thoughts @Akshay-2007-1 @martin-henz?

src/bundles/binary_tree/src/functions.ts
Akshay-2007-1: Never through the current call path, actually, the plugin wrapper in index.ts already throws before calling into this if no argument is passed, so value is guaranteed defined by the time it gets here. It's defensive coding for the fact that functions.ts is exported as a pure, evaluator-free API.

src/bundles/binary_tree/src/index.ts
Akshay-2007-1: fixed!

src/bundles/binary_tree/src/index.ts
Akshay-2007-1: Yeah, that'd be cleaner than the current "just don't declare a type" convention for predicates like is_tree. It touches Conductor itself though.

martin-henz: yes, is_tree should take ANY argument.

PR #791, opened 2026-07-13
Migrate midi module to Conductor

Akshay-2007-1:
## Description

Fixes #775

Migrates the `midi` module to be Conductor-ready.

`midi` is a dependency of both `sound` and `stereo_sound`, which import several of its functions (`midi_note_to_frequency`, `letter_name_to_midi_note`, `letter_name_to_frequency`) directly as plain TypeScript, and re-export them as part of their own Source-facing APIs. Rewriting `midi`'s functions to require an `IDataHandler` (the way `binary_tree`/`repeat` do) would have broken both of those call sites and re-exports immediately, so this migration keeps a pure, evaluator-free implementation available for cross-bundle TS consumption:

- `functions.ts` / `scales.ts` / `utils.ts` / `types.ts` — pure logic. `scales.ts` still builds js-slang-style lists via `pair`/`List`, same as before.
- `conductorAdapters.ts` — new, undecorated helper used only by the plugin: converts a js-slang scale list into a real Conductor list. Kept out of `index.ts` deliberately — a test file that imports `index.ts` directly hits a decorator syntax error under vitest's transform (it doesn't apply the same decorator settings `tsc` does), so anything meant to be unit-tested needs to live somewhere undecorated.
- `index.ts` — now a `BaseModulePlugin` subclass wrapping the pure functions. `SHARP`/`FLAT`/`NATURAL` are plain string constants, not functions, so `BaseModulePlugin.initialise()` (which only registers `exportedNames` that are functions) can't pick them up — they're pushed onto `this.exports` directly in the constructor instead.
- `sound`/`stereo_sound`: only their imports of `@sourceacademy/bundle-midi` changed, to `@sourceacademy/bundle-midi/functions` (the bundle root now exports the Conductor plugin instead of plain functions). No other changes to either bundle — their own Conductor migration is a separate, larger effort (they're ~90% duplicated implementations of each other; deferred pending a decision on whether to merge them).

Includes the same `__bindExportedMethods()` workaround `repeat`/`rune`/`binary_tree` use for the unbound-method issue in `BaseModulePlugin.initialise()` (source-academy/conductor#41, merged).

### `conductor-migration` had drifted from `master` for this bundle

While this migration was in flight, `master`'s `midi` picked up 6 new functions (`is_note_with_octave`, `add_octave_to_note`, `get_octave`, `get_note_name`, `get_accidental`, `key_signature_to_key`) and input validation on the existing ones (`midi_note_to_frequency` now range-checks its input), none of which existed on `conductor-migration`. This PR ports all of it, using the new `EvaluatorParameterTypeError`/`EvaluatorNumberRangeError`/`assertNumberWithinRange` from **source-academy/conductor#42** (a generic, reusable standardized-error-message addition to Conductor itself, not something bundle-specific) in place of `modules-lib`'s `InvalidParameterTypeError`/`assertNumberWithinRange`, which `functions.ts` can't depend on without pulling js-slang's `modules-lib` re-exports into `sound`/`stereo_sound`'s dependency graph transitively. Message format is unchanged — verified identical to `master`'s existing test expectations.

Also fixes `midi_note_to_letter_name`/`key_signature_to_key`'s `accidental` parameter to match `master`: it's the `Accidental` enum value (`'#'`/`'b'`), not the word `'flat'`/`'sharp'` my first pass used before finding the drift. `midi_note_to_letter_name` silently treats anything other than exactly `SHARP` as flat (matching `master`'s actual, slightly loose behavior) rather than validating it — `key_signature_to_key` is the one that validates, since its own switch has an explicit default case for that.

**Depends on source-academy/conductor#42** (not yet merged) for `EvaluatorParameterTypeError`/`EvaluatorNumberRangeError`/`assertNumberWithinRange`.

## Testing

- `yarn workspace @sourceacademy/bundle-midi run tsc` / `lint` / `test` (29/29) / `build` — all pass
- `yarn workspace @sourceacademy/bundle-sound run tsc` / `test` (28/28) — pass with the updated import path
- `yarn workspace @sourceacademy/bundle-stereo_sound run tsc` / `test` (18/18) — pass with the updated import path
- Verified against the published docs page (source-academy.github.io/modules/documentation/modules/midi.html): every documented function and constant (19 functions + 3 constants, no more, no less) confirmed working end-to-end through the actual compiled bundle, driving each export exactly the way a real evaluator calls a closure (detached, not bound to any instance) rather than just unit-testing the pure logic directly

### DIFF
--- a/src/bundles/midi/src/conductorAdapters.ts
+++ b/src/bundles/midi/src/conductorAdapters.ts
@@ -1,0 +1,21 @@
+/**
+ * Plain helpers used by the Conductor plugin in index.ts, kept in their own undecorated file so
+ * they stay importable from vitest — a test file that imports index.ts directly hits a decorator
+ * syntax error there (vitest's transform doesn't apply the same decorator settings tsc does).
+ */
+import { EvaluatorTypeError } from '@sourceacademy/conductor/common';
+import { DataType, type IDataHandler, type TypedValue } from '@sourceacademy/conductor/types';
+import { mEmptyList } from '@sourceacademy/conductor/util';
+import type { List } from 'js-slang/dist/stdlib/list';
+
+export async function scaleToConductorList(evaluator: IDataHandler, scale: List): Promise<TypedValue<DataType.LIST>> {
+  if (scale === null) return mEmptyList();
+  const [head, tail] = scale;
+  return evaluator.pair_make({ type: DataType.NUMBER, value: head }, await scaleToConductorList(evaluator, tail));
+}
+
+export function assertAccidental(value: string, funcName: string): asserts value is 'flat' | 'sharp' {
+  if (value !== 'flat' && value !== 'sharp') {
+    throw new EvaluatorTypeError(`${funcName} expects 'flat' or 'sharp'`, "'flat' | 'sharp'", `'${value}'`);
+  }
+}
--- a/src/bundles/midi/src/conductorAdapters.ts
+++ b/src/bundles/midi/src/conductorAdapters.ts
@@ -3,7 +3,6 @@
  * they stay importable from vitest — a test file that imports index.ts directly hits a decorator
  * syntax error there (vitest's transform doesn't apply the same decorator settings tsc does).
  */
-import { EvaluatorTypeError } from '@sourceacademy/conductor/common';
 import { DataType, type IDataHandler, type TypedValue } from '@sourceacademy/conductor/types';
 import { mEmptyList } from '@sourceacademy/conductor/util';
 import type { List } from 'js-slang/dist/stdlib/list';
@@ -12,10 +11,4 @@ export async function scaleToConductorList(evaluator: IDataHandler, scale: List)
   if (scale === null) return mEmptyList();
   const [head, tail] = scale;
   return evaluator.pair_make({ type: DataType.NUMBER, value: head }, await scaleToConductorList(evaluator, tail));
-}
-
-export function assertAccidental(value: string, funcName: string): asserts value is 'flat' | 'sharp' {
-  if (value !== 'flat' && value !== 'sharp') {
-    throw new EvaluatorTypeError(`${funcName} expects 'flat' or 'sharp'`, "'flat' | 'sharp'", `'${value}'`);
-  }
 }
--- a/src/bundles/midi/src/functions.ts
+++ b/src/bundles/midi/src/functions.ts
@@ -1,0 +1,130 @@
+/**
+ * Pure, evaluator-free implementations of the MIDI bundle's functions. Kept separate from the
+ * Conductor-facing `index.ts` plugin because other bundles (`sound`, `stereo_sound`) import these
+ * directly as plain TypeScript, not through Conductor's TypedValue/IDataHandler boundary.
+ *
+ * @module MIDI
+ * @author leeyi45
+ */
+
+import { Accidental, type MIDINote, type NoteWithOctave } from './types';
+import { midiNoteToNoteName, noteToValues } from './utils';
+
+/**
+ * Converts a letter name to its corresponding MIDI note.
+ * The letter name is represented in standard pitch notation.
+ * Examples are "A5", "Db3", "C#7".
+ * Refer to [this](https://i.imgur.com/qGQgmYr.png) mapping from
+ * letter name to midi notes.
+ *
+ * @param note given letter name
+ * @returns the corresponding midi note
+ * @example
+ * ```
+ * letter_name_to_midi_note('C4'); // Returns 60
+ * ```
+ * @function
+ */
+export function letter_name_to_midi_note(note: NoteWithOctave): MIDINote {
+  const [noteName, accidental, octave] = noteToValues(note, letter_name_to_midi_note.name);
+
+  let res = 12; // C0 is midi note 12
+  switch (noteName) {
+    case 'C':
+      break;
+    case 'D':
+      res += 2;
+      break;
+
+    case 'E':
+      res += 4;
+      break;
+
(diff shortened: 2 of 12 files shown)


src/bundles/midi/src/conductorAdapters.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Plain helpers used by the Conductor plugin in index.ts, kept in their own undecorated file so
+ * they stay importable from vitest — a test file that imports index.ts directly hits a decorator
+ * syntax error there (vitest's transform doesn't apply the same decorator settings tsc does).
+ */
+import { DataType, type IDataHandler, type TypedValue } from '@sourceacademy/conductor/types';
+import { mEmptyList } from '@sourceacademy/conductor/util';
+import type { List } from 'js-slang/dist/stdlib/list';
```
AaravMalani: Maybe should remove our dependency on `js-slang`? It's not important right now, but would be a nice to have

src/bundles/midi/src/index.ts
```diff
@@ -5,125 +5,201 @@
...
- * ```
- * letter_name_to_midi_note('C4'); // Returns 60
- * ```
- * @function
- */
-export function letter_name_to_midi_note(note: NoteWithOctave): MIDINote {
-  const [noteName, accidental, octave] = noteToValues(note, letter_name_to_midi_note.name);
+  @moduleMethod([DataType.CONST_STRING], DataType.NUMBER)
+  async* letter_name_to_midi_note(note: TypedValue<DataType.CONST_STRING>): AsyncGenerator<void, TypedValue<DataType.NUMBER>, unknown> {
+    return { type: DataType.NUMBER, value: letter_name_to_midi_note_func(note.value as NoteWithOctave) };
```
AaravMalani: I disagree with the function signature of these functions; if they can accept any string and throw errors for invalid strings, they should accept strings and they shouldn't need the cast.

src/bundles/midi/src/functions.ts
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Converts the key signature to the corresponding key
+ * @function
+ * @example
+ * ```
+ * key_signature_to_key(SHARP, 2); // Returns "D", since the key of D has 2 sharps
+ * key_signature_to_key(FLAT, 3); // Returns "Eb", since the key of Eb has 3 flats
+ * ```
+ */
+export function key_signature_to_key(accidental: Accidental.FLAT | Accidental.SHARP, numAccidentals: number): Note {
```
AaravMalani: This is unrelated to the migration and my circle of fifths is rusty, but doesn't every major key have a relative minor (D <-> Bm)? Shouldn't it be made clear in the documentation that it returns the major key / have a way to return it in a scale of the user's choosing?

AaravMalani: Actually, all the scales are modes of the major scale, so technically we should be able to support it returning the key in any of the scales (I'm surprised natural and harmonic minor are missing from the scales)

src/bundles/midi/src/index.ts
```diff
@@ -5,125 +5,201 @@
...
+    'lydian_scale',
+    'mixolydian_scale',
+    'minor_scale',
+    'aeolian_scale',
+    'locrian_scale'
+  ] as const;
+  static channelAttach = [];
+  constructor(conduit: IConduit, channels: IChannel<any>[], evaluator: IInterfacableEvaluator) {
+    super(conduit, channels, evaluator);
+    this.__bindExportedMethods();
```
AaravMalani: Removable now (thanks!)

src/bundles/midi/src/conductorAdapters.ts
Akshay-2007-1: Agreed, this is really just importing a two-element tuple type for List/Pair, nothing js-slang-specific. Can swap it for a small local type alias so midi doesn't need js-slang as a dependency at all.

src/bundles/midi/src/index.ts
Akshay-2007-1: fixed!

src/bundles/midi/src/functions.ts
Akshay-2007-1: @gemini-code-assist, has this been fixed?

Akshay-2007-1: @AaravMalani you're right that all these scales are modes of the same major scale under the hood. But, this behavior's carried over unchanged from the original js-slang module though. @martin-henz, any thoughts? Should we leave it be?

src/bundles/midi/src/index.ts
Akshay-2007-1: fixed!

src/bundles/midi/src/conductorAdapters.ts
Akshay-2007-1: @gemini-code-assist, I believe this has been fixed in the latest commit?

src/bundles/midi/src/functions.ts
leeyi45: Yes every major key has a relative minor.

Natural minor is included, I presume @AaravMalani you're referring to the harmonic and melodic minor scales.
The thing is that there's a lot of scales out there (like major and minor pentatonic), we have to draw the line somewhere.

For `key_signature_to_key` the assumption is that it returns the major scale with that key signature. I guess we should update the documentation to reflect that.

martin-henz: oh, let's not worry about consistency with musical theory at this point. Let's aim for feature parity. If there is time, we can do improvements once everything works, or do an overhaul as part of CP3108.

